PR #9, opened 2017-07-14
proto instance + misc. fixes

junmipan:
SPOPT-1329


junmipan:
This is the last branch I create on ligato.  Will fork for future dev.

PR #40, opened 2017-08-04
Cassandra struct exported fields & table/schema customization using methods

jozef-slezak:


utils/structs/structs_reflection.go
```diff
@@ -44,36 +44,103 @@ func FindField(pointerToAField interface{}, pointerToAStruct interface{}) (field
...
+}
+
+// ListExportedFieldsPtrs iterates struct fields and return slice of pointers to field values
+func ListExportedFieldsPtrs(val interface{}, predicates ...ExportedPredicate) []interface{} {
+	rVal := reflect.Indirect(reflect.ValueOf(val))
+	ptrs := []interface{}{}
+	for i := 0; i < rVal.NumField(); i++ {
+		field := rVal.Field(i)
+		structField := rVal.Type().Field(i)
+		if !FieldExported(&structField) {
```
milanlenco: What about: if !FieldExported(&structField, predicates...) {
Then we can avoid the for cycle below to appear in this function as well as in FieldExported().

PR #45, opened 2017-08-08
Editorial changes in README and guidelines files.

jmedved:


docs/guidelines/PLUGIN_FLAVOURS.md
```diff
@@ -25,33 +32,33 @@ import (
...
-	Basic    ReusedFlavour
+type CompositeFlavor struct {
+	Basic    ReusedFlavor
 	PluginXY PluginXY
 	injected bool
 }
 
-func (flavour *CompositeFlavour) Inject() error {
-	if flavour.injected {
+func (Flavor *CompositeFlavor) Inject() error {
```
jozef-slezak: accidently renamed also instance name

docs/guidelines/PLUGIN_DEPENDENCIES.md
```diff
@@ -14,10 +14,12 @@
 ```
 	
 2. For plugins, constructors are not needed. The reasons:
-  * the dependencies are supposed to be exported fields (and injected).
-  * Init() method is called on the plugin during agent startup [(see StartAgent in example main() function)(../../examples/simple-agent)
+  * The dependencies are supposed to be the exported fields (and injected).
+  * The Init() method is called on the plugin during agent startup; 
+    see []StartAgent in the example main() function the
```
jozef-slezak: do we really need to change the hyperlink?

PR #48, opened 2017-08-09
Editorial changes to README files.

jmedved:


docs/guidelines/SYSTEM_INTEGRATION.md
```diff
@@ -15,18 +16,24 @@ TODO link to code of db or messaging that allows to configure global timeout
...
 TODO link to code/doc of db or messaging 
 ```
 
-# AfterInit() failed to connect
-If it is not able to connect during timeout, the plugin needs to propagate errors. The Agent will not start. 
-TODO assuming that there is a default deployment strategy for container-based cloud (as is with K8s) that 
-will try to heal the container and basically recreates it.
+## AfterInit() failed to connect to a service
+If a plugin implements a client that consumes an external service, and
+the client is unable to connect to hte service before a timeout, the plugin
```
lukasmacko: the service

examples/README.md
```diff
@@ -1,15 +1,21 @@
...
-* **[flags](flags/main.go)** example registers flags and shows their runtime values
-* **[kafka](kafka/main.go)** creates simple plugin which registers Kafka consumer and sends a test notification
-* **[logs](logs/main.go)** shows the logger and log level usage
+* **[etcd](etcd/main.go)** uses the ETCD data broker to write data into 
+  ETCD, which are then caught by the watcher
+* **[flags](flags/main.go)** example registers flags and shows their 
+  runtime values
+* **[kafka](kafka/main.go)** creates a simple plugin which registers a 
+  Kafka consumer and sends a test notification
+* **[logs](logs/main.go)** shows how to use the logger and wotk wiht
```
lukasmacko: work with

PR #49, opened 2017-08-10
ODPM-467: http probe in separate port - impl

VladoLavor:
- custom probe port can be set in its own flag


httpmux/http_plugin_impl.go
```diff
@@ -2,41 +2,50 @@ package httpmux
...
-const PluginID core.PluginName = "HTTP"
-
 const (
-	// DefaultHTTPPort is used during HTTP server startup unless different port was configured
+	// PluginID used in the Agent Core flavors
+	PluginID core.PluginName = "HTTP"
+	// DefaultHTTPPort is used as a default HTTP rest server startup unless different port was configured
 	DefaultHTTPPort = "9191"
+	// DefaultProbePort is used as a default HTTP probe server startup unless different port was configured
+	DefaultProbePort = "9190"
```
jozef-slezak: Please move this flag to generic flavor (we do not want to uglify the httpmux based on statuscheck related requirements). Then inject the value to httpmux. There should be also logic in generic flavor that compares if probe port & http port is different. Keep the values same please by default. If the values are same then do not create new instance of httpmux but if it is different inject new instance.

Two more extra work items:
1. create dedicated structure for httpmux configuration with string endpoint istead of http port. This new structure will replace HTTPPort injectable field.
2.  move status-check to health/status-check. By doing this we have consistent diagram in root README.md with package hierarchy and we are ready to extract liveness & readiness plugin from status-check.

flavors/generic/generic_flavor.go
```diff
@@ -40,11 +42,15 @@ func (f *FlavorGeneric) Inject() error {
...
 	f.HTTP.LogFactory = &f.Logrus
-	//TODO f.HTTP.Logger = f.Logrus.LoggerWithPrefix(f.PluginName(&f.HTTP))
-	//TODO f.HTTP.Config = f.Config.ConfigWithPrefix(f.PluginName(&f.HTTP))
+	f.HTTP.HTTPport = &f.Probe.HTTPPort
+	//TODO f.HTTPProbe.Logger = f.Logrus.LoggerWithPrefix(f.PluginName(&f.HTTPProbe))
+	//TODO f.HTTPProbe.Config = f.Config.ConfigWithPrefix(f.PluginName(&f.HTTPProbe))
 	f.LogManager.ManagedLoggers = &f.Logrus
 	f.LogManager.HTTP = &f.HTTP
+	f.StatusCheck.LogFactory = &f.Logrus
+	f.StatusCheck.Probe = &f.Probe.ProbePort
```
jozef-slezak: rename ProbePort to HTTPPort

health/probe/probes.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// Default port used for http and probing
+const defaultPort = "9191"
+
+// PluginID used in the Agent Core flavors
+const PluginID core.PluginName = "Probe"
+
+var (
+	httpPort string
```
jozef-slezak: this seems to be obsolete

PR #62, opened 2017-08-15
Update doc.go

martinfrk:
minor edits (grammar & typos)


statuscheck/doc.go
```diff
@@ -12,17 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package statuscheck provides the API for reporting status changes from plugins to the statuscheck plugin,
-// that exposes it via ETCD and HTTP.
+// The statuscheck package provides an API for reporting status changes from plugins to the statuscheck plugin,
```
lukasmacko: This can't be changed because of golint (static analysis tool). It requires the package level comment to start with word `Package`...

martinfrk: Understood. We can keep it as it is (and in other similar instances as well).

PR #75, opened 2017-08-16
Improve naming in the idxmap

lukasmacko:


jmedved:
The only thing i would point out is that the granularity of subscription seems to be per-plugin. That means, I can subscribe to all mappings from a plugin, or none. What if a plugin has defined two mappings, and i am interested in only one of them?

I am merging this pull request, but the above is something to consider in a future update.

PR #77, opened 2017-08-16
Added error handling to iterators

junmipan:


db/keyval/redis/bytes_broker_impl.go
```diff
@@ -220,16 +229,24 @@ func (it *bytesKeyIterator) GetNext() (key string, rev int64, lastReceived bool)
 	return key, 0, false
 }
 
+func (it *bytesKeyValIterator) Close() error {
```
jozef-slezak: Could you please document the behaviour about errors?

junmipan: Yes, I will.  In line or separate doc?

jozef-slezak: Yes, I mean Close() method comment. There should be also "DISCLAIMER" also in method that returns iterator...

PR #84, opened 2017-08-17
doc: iterator behavor

junmipan:


junmipan:
Please note that the Close() method has not yet been added to db/keyval/*Iterator interfaces.  The change will have broad impact on API's currently implementing these interfaces.
On the other hand, the iterators in redis API do implement Close() methods.  So when the db/keyval/*Iterator interfaces are enhanced, the redis API will automatically adapt.

PR #89, opened 2017-08-22
ODPM-361 flavors/localdeps: Log+PluginName, PluginInfraDeps

jozef-slezak:


datasync/resync/registration.go
```diff
@@ -14,8 +14,11 @@
 
 package resync
 
+import "fmt"
+
 // Registration in an interface that is returned by the Register() call
```
lukasmacko: is

PR #92, opened 2017-08-22
introduced PluginConfig and moved package util/config to /config

jozef-slezak:


config/plugin_config.go
```diff
@@ -0,0 +1,46 @@
+package config
+
+import (
+	"flag"
+)
+
+// Config is API for plugins to access configuration.
```
rastislavszabo: this will report a golint issue once we add config/ dir for scanning.. it should be PluginConfig...

PR #137, opened 2017-09-08
SPOPT-1569 - health status for Cassandra & Redis plugins

mpundlik:


db/sql/cassandra/plugin_impl_cassa.go
```diff
@@ -15,11 +15,20 @@
...
+	"github.com/gocql/gocql"
+	"github.com/ligato/cn-infra/core"
 	"github.com/ligato/cn-infra/db/sql"
 	"github.com/ligato/cn-infra/flavors/local"
+	"github.com/ligato/cn-infra/health/statuscheck"
 	"github.com/willfaught/gockle"
 )
 
+// HealthCheck is a structure used to represent table to help verify health status
+type HealthCheck struct {
```
jozef-slezak: Just do some read-only query using broker.Exec(). It is too much to have this not needed table on a particular schema.

db/sql/cassandra/plugin_impl_cassa.go
```diff
@@ -73,18 +82,28 @@ func (p *Plugin) AfterInit() error {
 		p.session = gockle.NewSession(session)
 	}
 
-	/* TODO Register for providing status reports (polling mode)
+	// Register for providing status reports (polling mode)
 	if p.StatusCheck != nil && p.session != nil {
 		p.StatusCheck.Register(core.PluginName(p.String()), func() (statuscheck.PluginState, error) {
-			_, _, err := p.Skeleton.NewBroker("/").GetValue(healthCheckProbeKey, nil)
+			broker := p.NewBroker()
```
jozef-slezak: just do broker.Exec("readonly cassandra query that is runalbe in every schema without having extra objects")

mpundlik: done

examples/redis_plugin/main.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Redis flag to load config
+func init() {
+	flag.String("redis-config", "redis.conf",
+		"Location of the redis configuration file")
+}
+
+// ExampleFlavor is a set of plugins required for the redis example.
+type ExampleFlavor struct {
```
jozef-slezak: could you please move the example flavor & Deps struct to dedicated file inside example dir (Milan is currently doing that for all other examles)? thanks

db/keyval/redis/plugin_impl_redis.go
```diff
@@ -57,6 +64,28 @@ func (p *Plugin) Init() error {
...
+// AfterInit is called by the Agent Core after all plugins have been initialized.
+func (p *Plugin) AfterInit() error {
+	if p.disabled {
+		return nil
+	}
+
+	// Register for providing status reports (polling mode)
+	if p.StatusCheck != nil {
+		p.StatusCheck.Register(core.PluginName(p.String()), func() (statuscheck.PluginState, error) {
+			_, _, err := p.Skeleton.NewBroker("/").GetValue(healthCheckProbeKey, nil)
```
jozef-slezak: If possbile try to use directly connection broker to not create new instance again

db/sql/cassandra/plugin_impl_cassa.go
jozef-slezak: If possbile try to use directly connection broker to not create new instance again

mpundlik: done

examples/redis_plugin/main.go
mpundlik: done

PR #147, opened 2017-09-14
ODPM 552: No need to have status model under another sub-directory.

milanlenco:


health/probe/plugin_impl_probes.go
```diff
@@ -22,7 +22,7 @@ import (
 	"github.com/ligato/cn-infra/core"
 	"github.com/ligato/cn-infra/flavors/local"
 	"github.com/ligato/cn-infra/health/statuscheck"
-	"github.com/ligato/cn-infra/health/statuscheck/model/status"
+	status "github.com/ligato/cn-infra/health/statuscheck/model"
```
jozef-slezak: I do not like this change because it brings problems while coding. I understand that we want to have minimal depth in directory structure...

Problem is while importing this package because we would introduce package aliases. I think this change would be against GO conventions.
GO conventions prohibit to create package utils. My understanding is that that it is wrong to have utils/file.go but it is ok to have utils/package/file.go. There is analogy between model and utils. 

Githubs also handles this correctly while browsing the repository.
Godoc would handle that correctly if we create doc.go.

Bottom lines: let's avoid import aliases whenever possible.

rastislavszabo: I agree with Jozef on this..

jozef-slezak:
I see following options/alternatives:
1. health/statuscheck/model/status -> health/statuscheck/statusmodel (and update conventions...)
2. health/statuscheck/model/status -> health/statuscheck/status
3. leave it as it is health/statuscheck/model/status (my prefference)

Jan, please let us know your preference.

PR #153, opened 2017-09-19
ODPM-586: removed runtime errors from safeclose

VladoLavor:


db/sql/cassandra/plugin_impl_cassa.go
```diff
@@ -120,7 +120,9 @@ func (p *Plugin) NewBroker() sql.Broker {
 
 // Close resources
 func (p *Plugin) Close() error {
-	p.session.Close()
+	if p.session != nil {
```
jozef-slezak: just call safeclose

PR #155, opened 2017-09-19
ODPM-569: plugin deps replaced with interfaces

VladoLavor:

### DIFF
--- a/datasync/msgsync/plugin_impl_msgsync.go
+++ b/datasync/msgsync/plugin_impl_msgsync.go
@@ -21,7 +21,6 @@ import (
 	"github.com/ligato/cn-infra/datasync"
 	"github.com/ligato/cn-infra/flavors/local"
 	"github.com/ligato/cn-infra/messaging"
-	"github.com/ligato/cn-infra/messaging/kafka"
 )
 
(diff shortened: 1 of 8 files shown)


health/probe/plugin_impl_probes.go
```diff
@@ -75,11 +75,7 @@ func (p *Plugin) Init() (err error) {
 				HTTPport: httpPort,
 			},
 		}
-		err := p.HTTP.Init()
-		if err != nil {
-			return err
-		}
-		err = p.HTTP.AfterInit()
+		err := p.HTTP.ExternalInit()
```
jozef-slezak: No need to do that - just rebase. Yesterday we have cleaned up this - basically removed nested plugin initialization.

datasync/msgsync/plugin_impl_msgsync.go
```diff
@@ -35,7 +34,7 @@ type PubPlugin struct {
 // to not mix with other plugin fields.
 type Deps struct {
 	local.PluginInfraDeps               // inject
-	Messaging             *kafka.Plugin // inject
+	Messaging             messaging.Mux // inject
```
jozef-slezak: Please, check also the examples (replacement of pointers by interfaces needs to be done also in there).

PR #161, opened 2017-09-21
ODPM-610: kafka enhancements

VladoLavor:


messaging/kafka/client/syncproducer.go
```diff
@@ -124,10 +118,12 @@ func (ref *SyncProducer) Close() error {
 	}
 	ref.Debug("SyncProducer closed")
 
-	err = ref.Client.Close()
-	if err != nil {
-		ref.Errorf("client close error: %v", err)
-		return err
+	if !ref.Client.Closed() {
```
lukasmacko: I don't think it's a good idea to close always close the client. It might still be used somewhere else.
I would suggest to close the client only if it was create in the `New*Producer` function.
This applies for both producer types.

messaging/kafka/client/syncproducer.go
```diff
@@ -37,8 +37,9 @@ type SyncProducer struct {
 	sync.Mutex
 }
 
-// NewSyncProducer returns a new SyncProducer
-func NewSyncProducer(config *Config, wg *sync.WaitGroup) (*SyncProducer, error) {
+// NewSyncProducer returns a new SyncProducer instance. Producer is created from provided sarama client. Also
+// the partitioner is set here. Note: sarama partitioner should match the one used in config.
+func NewSyncProducer(config *Config, sClient sarama.Client, partitioner string, wg *sync.WaitGroup) (*SyncProducer, error) {
```
lukasmacko: Please, either handle in this function case when sClient is nil and create own new client that will be used or rename this one to `NewSyncProducerFromClient` and handle the case in separate function.

See also the comment about closing the client.

messaging/kafka/plugin_impl_kafka.go
```diff
@@ -50,154 +52,170 @@ type Deps struct {
...
 	}
 
 	// Register for providing status reports (polling mode)
-	if p.StatusCheck != nil {
-		p.StatusCheck.Register(p.PluginName, func() (statuscheck.PluginState, error) {
+	if plugin.StatusCheck != nil {
+		plugin.StatusCheck.Register(plugin.PluginName, func() (statuscheck.PluginState, error) {
 			// Method 'RefreshMetadata()' returns error if kafka server is unavailable
-			err := p.consumer.Client.RefreshMetadata(topic)
+			err := plugin.consumer.Client.RefreshMetadata(topic)
```
lukasmacko: call directly plugin.hsClient.RefreshMetadata. the consumer is unnecessary here.

messaging/kafka/plugin_impl_kafka.go
```diff
@@ -50,168 +52,170 @@ type Deps struct {
...
-		}
-		p.Log.Debug("Default multiplexer initialized")
+	// init 'manual' sarama client
+	plugin.manClient, err = client.NewClient(clientCfg, client.Manual)
+	if err != nil {
+		return err
+	}
+
+	// init bsm/sarama consumer
+	plugin.consumer, err = client.NewConsumer(clientCfg, true,nil)
```
lukasmacko: is this used anywhere? I think it can be removed

PR #169, opened 2017-09-26
SPOPT-1621 - Prometheus/Telemetry health status

mpundlik:
Contains health status changes and fix for Cassandra queries


health/probe/plugin_impl_prometheus.go
```diff
@@ -124,6 +121,61 @@ func (p *PrometheusPlugin) AfterInit() error {
...
+			p.registerGauge(
+				Namespace,
+				Subsystem,
+				DependencyHealthName,
+				DependencyHealthHelp,
+				prometheus.Labels{
+					ServiceLabel:    p.getServiceLabel(),
+					DependencyLabel: k,
+				},
+				p.getDependencyHealth(k, v),
```
junmipan: Plugin status 'v' should not be collected here.  Rather, should be collected when getDependencyHealth() is invoked -- i.e., inside its body.

PR #177, opened 2017-09-29
SPOPT-1621, 1658 - improvements for Prometheus metrics and vpp-agent

mpundlik:


health/probe/plugin_impl_prometheus.go
```diff
@@ -121,60 +121,26 @@ func (p *PrometheusPlugin) AfterInit() error {
...
 				Subsystem,
 				DependencyHealthName,
 				DependencyHealthHelp,
 				prometheus.Labels{
 					ServiceLabel:    p.getServiceLabel(),
-					DependencyLabel: k,
+					DependencyLabel: v,
 				},
-				p.getDependencyHealth(k, v),
+				p.getDependencyHealth(v, pluginStatusIdx),
```
jozef-slezak: statuscheck allows to notify when plugin status changes using watch (instead of polling in getDependencyHealth) is it technically possible with prometheus?

mpundlik: Would be good if we could have a quick discussion.

PR #185, opened 2017-10-02
Kafka tests - added Kafka post-init-consumer, improvement for assertion grep 

stano29:


scripts/test_examples.sh
```diff
@@ -201,44 +201,457 @@ stopKafka
...
+  -messageCount=\"\": Number of messages which will be send. Set to '0' to just watch
+  -microservice-label=\"vpp1\": microservice label; also set via 'MICROSERVICE_LABEL' env variable.
+  -offsetMsg=\"\": Use 'latest', 'oldest' or exact number of message offset
+")
+
+cmd="examples/kafka-plugin/manual-partitioner/manual-partitioner --kafka-config examples/kafka-plugin/manual-partitioner/kafka.conf -h"
+testOutput "${cmd}" "${expected}"
+
+# Let us test the running without parameters - in example are generated 10 Kafka Messages to both topics but the consumed is only 5 messages from each topic beginning with offset 5
+expected=("Invalid partition -1 - defaulting to 0
```
VladoLavor: please do not expect log messages which are not called in the example main.go. Check everywhere

scripts/test_examples.sh
```diff
@@ -201,44 +201,457 @@ stopKafka
...
+Received sync Kafka Message, topic 'example-sync-topic', partition '1', offset '7', key: 'proto-key', 
+Received sync Kafka Message, topic 'example-sync-topic', partition '1', offset '8', key: 'proto-key', 
+Received sync Kafka Message, topic 'example-sync-topic', partition '1', offset '9', key: 'proto-key', 
+Sending 10 async Kafka notifications (protobuf) ...
+Async message successfully delivered
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '5', key: 'async-proto-key', 
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '6', key: 'async-proto-key', 
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '7', key: 'async-proto-key', 
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '8', key: 'async-proto-key', 
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '9', key: 'async-proto-key',
```
VladoLavor: you are missing Sync watcher closed and Async watcher closed

scripts/test_examples.sh
```diff
@@ -201,44 +201,457 @@ stopKafka
...
+# Let us test - in example are generated next more 10 Kafka Messages to both topics
+expected=("Invalid partition -1 - defaulting to 0
+offset arg not set, using default value
+messageCount arg not set, using default value
+Offset: 0, message count: 10
+Started Kafka sync event handler...
+Started Kafka async event handler...
+All plugins initialized successfully
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '5', key: 'async-proto-key', 
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '6', key: 'async-proto-key',
```
VladoLavor: keep it short, 
Received async Kafka Message, topic 'example-async-topic', partition '2', offset '5', key: 'async-proto-key',
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '9', key: 'async-proto-key',
will be enough

scripts/test_examples.sh
```diff
@@ -201,44 +201,457 @@ stopKafka
...
+#time="2017-09-29 14:23:04.54780" level=info msg="Agent AfterInit took 0s" durationInNs=0 loc="core/agent_core.go(124)" logger=core tag=00000000 
+#time="2017-09-29 14:23:04.54782" level=error msg="Error loading core: plugin KafkaExample: Init error ''messageCount' has to be a number, not "0"', took 44.642�s" loc="core/event_loop.go(28)" logger=core tag=00000000 
+#Testing examples/kafka-plugin/manual-partitioner/manual-partitioner --kafka-config examples/kafka-plugin/manual-partitioner/kafka.conf -messageCount="0"
+
+#workaround
+cmd='examples/kafka-plugin/manual-partitioner/manual-partitioner --kafka-config examples/kafka-plugin/manual-partitioner/kafka.conf --messageCount 0'
+testOutput "${cmd}" "${expected}"
+
+
+# Let us test - in example one new message generated
```
VladoLavor: It seems this test does the same as the one above, i think we can keep just one of them

scripts/test_examples.sh
```diff
@@ -201,44 +201,457 @@ stopKafka
...
+Async message successfully delivered
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '21', key: 'async-proto-key', 
+")
+
+
+cmd="examples/kafka-plugin/manual-partitioner/manual-partitioner --kafka-config examples/kafka-plugin/manual-partitioner/kafka.conf --messageCount 1 --offsetMsg 18"
+testOutput "${cmd}" "${expected}"
+
+
+# Let us test - in example no new message generated
```
VladoLavor: again - this test is the same as the one on line 399

scripts/test_examples.sh
```diff
@@ -201,44 +201,457 @@ stopKafka
...
 startKafka
 
-expected=("Sending 10 sync Kafka notifications
-Sending 10 async Kafka notifications
-Received Kafka Message, topic 'example-sync-topic'
-Async message successfully delivered, topic 'example-async-topic'
-Received async Kafka Message, topic 'example-async-topic'
-Sync watcher closed
-Async watcher closed
+# Let us test parameters display
```
VladoLavor: I'm thinking whether this have some added value. The goal of example testing is to verify kafka functionality. This has nothing to do with it.

scripts/test_examples.sh
stano29: The results of both discussed are same, but the first one is called without parameter and generating new messages and the second one is called with parameter --messageCount 0 - only lists the messages which were generated to that point.

I suggest to shorten the list of asserted messages and leave this second case.

scripts/test_examples.sh
stano29: I removed all the parameters tests instances...

scripts/test_examples.sh
```diff
@@ -8,15 +8,16 @@ PREV_IFS="$IFS"
 # arguments
```
VladoLavor: please do not commit the scripts/test_examples.sh.7z binary

scripts/test_examples.sh
```diff
@@ -201,43 +212,255 @@ stopKafka
...
+Sending 0 sync Kafka notifications (protobuf) ...
+Sending 0 async Kafka notifications (protobuf) ...
+Sync watcher closed
+Async watcher closed
+")
+
+unexpected=("Error while stopping watcher
+")
+
+#this is not working - I do not know why
```
VladoLavor: try  #cmd="examples/kafka-plugin/manual-partitioner/manual-partitioner --kafka-config examples/kafka-plugin/manual-partitioner/kafka.conf -messageCount=0" or remove it

scripts/test_examples.sh
```diff
@@ -201,43 +212,255 @@ stopKafka
...
+Sending 1 sync Kafka notifications (protobuf) ...
+Received sync Kafka Message, topic 'example-sync-topic', partition '1', offset '11', key: 'proto-key', 
+Sending 1 async Kafka notifications (protobuf) ...
+Async message successfully delivered, topic 'example-async-topic', partition '2', offset '11', key: 'async-proto-key'
+Received async Kafka Message, topic 'example-async-topic', partition '2', offset '11', key: 'async-proto-key', 
+Sync watcher closed
+Async watcher closed
+")
+
+unexpected=("Error while stopping watcher
```
VladoLavor: add here Received async Kafka Message, topic 'example-async-topic', partition '2', offset '7', key: 'async-proto-key',
since we are watching offset 8, we do not want to see the offset before it.

scripts/test_examples.sh
```diff
@@ -201,43 +212,255 @@ stopKafka
...
 #### Kafka-plugin hash-partitioner #######################################
 
 startKafka
 
-expected=("Sending 10 sync Kafka notifications
-Sending 10 async Kafka notifications
-Received Kafka Message, topic 'example-sync-topic'
-Async message successfully delivered, topic 'example-async-topic'
-Received async Kafka Message, topic 'example-async-topic'
+# Let us test the running without parameters - in example are generated 10 Kafka Messages to both topics
```
VladoLavor: in every hash partitioner example - consider making of partition number wildcarded. Hash partitioner is choosing the partitions according to the key (async-proto-key for example, that means all sync messages should be on the same partition and all async messages should be on a different partition). Although partition numbers are the same in most cases, there can be rare cases when the partition numbers will be different and the test will fail

scripts/test_examples.sh
stano29: Done

scripts/test_examples.sh
stano29: At the present time the hash-partitioner example uses kafka docker started with unchanged server.properties where there is the setting:
# The default number of log partitions per topic
num.partitions=1

So maybe it could be better to add the other test case where num.partitions per topic is 3 and in this case it makes sense to make partition number wildcarded?!

stano29:
At the present time the hash-partitioner example uses kafka docker started with unchanged server.properties where there is the setting:
num.partitions=1
So maybe it could be better to add the other test case where num.partitions per topic is 3 and in this case it makes sense to make partition number wildcarded?!

PR #186, opened 2017-10-03
Expose delete in kvdbsync plugin

lukasmacko:


lukasmacko:
Note: I've not added the `Delete` function into `KeyProtoValWriter` since delete is not supported in kafka that implements this interface as well.

PR #189, opened 2017-10-05
ODPM-440: Update github.com/onsi/gomega to v1.2.0

ondrej-fabry:
Updating github.com/onsi/gomega dependency to v1.2.0 to match version for vpp-agent.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -64,7 +64,7 @@ import:
   version: 2.1.0
 - package: github.com/howeyc/crc16
 - package: github.com/onsi/gomega
-  version: v1.1.0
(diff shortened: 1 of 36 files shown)


glide.yaml
```diff
@@ -64,7 +64,7 @@ import:
   version: 2.1.0
 - package: github.com/howeyc/crc16
 - package: github.com/onsi/gomega
-  version: v1.1.0
+  version: v1.2.0
```
jozef-slezak: you need to update also vendor probably

ondrej-fabry: Pushed with updated gomega in vendor

PR #192, opened 2017-10-05
ODPM-440: Add some missing dependencies to glide.yaml update in vendor

ondrej-fabry:
Packages added to `glide.yaml`:

- gopkg.in/yaml.v2
- github.com/mattn/go-isatty
- github.com/davecgh/go-spew
- github.com/gogo/protobuf


glide.yaml
```diff
@@ -69,3 +69,13 @@ import:
   version: v0.8.0
   subpackages:
   - prometheus/promhttp
+- package: gopkg.in/yaml.v2
```
VladoLavor: This dependency is defined in glide.yaml in coreos/etcd with version cd8b52f8269e0feb286dfeef29f8fe4d5b397e0b (the same I got when I tried) and we definitely should not change the version here. Glide flattens this package to the same version every time so I think this does not need to be here. 

Others seem ok to me

ondrej-fabry: Ok, removed it

PR #227, opened 2017-12-01
ODPM-642: add interface state to readiness probe 

VladoLavor:


health/probe/plugin_impl_probes.go
```diff
@@ -59,14 +59,16 @@ func (p *Plugin) AfterInit() error {
...
+		agentStat.InterfaceStatus = &ifStat
+		agentStatJSON, _ := json.Marshal(agentStat)
+		if agentStat.State == status.OperationalState_OK {
 			w.WriteHeader(http.StatusOK)
-			w.Write(statJSON)
+			w.Write(agentStatJSON)
 		} else {
 			w.WriteHeader(http.StatusInternalServerError)
-			w.Write(statJSON)
+			w.Write(agentStatJSON)
```
ondrej-fabry: The `w.Write(agentStatJSON)` part can be outside of *if .. else ..*

health/statuscheck/model/status/status.proto
```diff
@@ -24,3 +25,14 @@ message PluginStatus {
     int64 last_update = 5;  /* last update of the state */
     string error = 6;       /* last seen error */
 }
+
+message InterfaceStatus {
+    message Interfaces {
```
ondrej-fabry: This should be `Interface`.

health/statuscheck/model/status/status.proto
```diff
@@ -24,3 +25,14 @@ message PluginStatus {
     int64 last_update = 5;  /* last update of the state */
     string error = 6;       /* last seen error */
 }
+
+message InterfaceStatus {
```
ondrej-fabry: This should include plural in the name.

health/statuscheck/plugin_impl_statuscheck.go
```diff
@@ -73,6 +75,10 @@ func (p *Plugin) Init() error {
 		LastChange:   time.Now().Unix(),
 	}
 
+	// initial empty interface status
+	var statusData []*status.InterfaceStatus_Interfaces
+	p.interfaceStat = &status.InterfaceStatus{Interfaces: statusData}
```
ondrej-fabry: You can omit setting the field `Interfaces` since it will be `nil` by default anyway.

health/statuscheck/plugin_impl_statuscheck.go
```diff
@@ -195,6 +217,36 @@ func (p *Plugin) ReportStateChange(pluginName core.PluginName, state PluginState
 	}
 }
 
+func (p *Plugin) reportInterfaceStateChange(data *status.InterfaceStatus_Interfaces) {
+	// Filter interfaces without internal name
```
ondrej-fabry: This doesn't synchronize access with `p.access` and might cause race condition on multiple calls to `ReportStateChangeWithMeta`.

health/statuscheck/model/status/status.proto
```diff
@@ -24,3 +25,14 @@ message PluginStatus {
...
+
+message InterfaceStats {
+    message Interface {
+        string internal_name = 1;          /* interface name used in VPP */
+        uint32 index = 3;                  /* sw interface index */
+        string status = 4;                 /* status (UP, DOWN, DELETED, ...) */
+        string ip_address = 5;
+        string mac_address = 6;
+    }
+    repeated Interface interface = 1;
```
ondrej-fabry: This should be `repeated Interface

PR #231, opened 2017-12-14
remove unwanted transport adapter check

VladoLavor:

Transport aggregator returns an error in case there is no transport available.  With the possibility of omitting to send status data, this error is no longer reasonable.  

@ondrej-fabry please add this change to the vpp-agent


datasync/aggregator.go
```diff
@@ -66,9 +66,6 @@ func (ta *CompositeKVProtoWatcher) Watch(resyncName string, changeChan chan Chan
 // Put writes data to all aggregated transports.
 // This function implements KeyProtoValWriter.Put().
 func (ta *CompositeKVProtoWriter) Put(key string, data proto.Message, opts ...PutOption) error {
-	if len(ta.Adapters) == 0 {
-		return fmt.Errorf("no transport is available in aggregator")
```
ondrej-fabry: This makes `fmt` import unused.

PR #268, opened 2018-05-21
Handle SIGTERM in EventLoopWithInterrupt

mestery:
This handles SIGTERM in addition to SIGINT. Relevant for the cases where the
running program is sent SIGTERM, such as when running as a daemonset.

### DIFF
--- a/core/event_loop.go
+++ b/core/event_loop.go
@@ -17,11 +17,12 @@ package core
 import (
 	"os"
 	"os/signal"
+	"syscall"
 )
 
 // EventLoopWithInterrupt starts an instance of the agent created with NewAgent().
-// Agent is stopped when <closeChan> is closed or a user interrupt (SIGINT)
-// is received.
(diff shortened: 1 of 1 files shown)


core/event_loop.go
```diff
@@ -31,6 +32,7 @@ func EventLoopWithInterrupt(agent *Agent, closeChan chan struct{}) error {
 
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, os.Interrupt)
```
mestery: Some of the examples I've using `signal.Notify()` I've seen use sys call.SIGINT instead of os.Interrupt. I suspect they are the same, but for consistency, I thought I'd mention this as well.

mestery:
@ondrej-fabry Would be curious to get your input on this one.

core/event_loop.go
ondrej-fabry: It's essentially the same thing: https://golang.org/src/os/exec_posix.go?s=552:585#L18

PR #273, opened 2018-05-24
Allow configuration of grpc socket type

edwarnicke:
gRPC allows "unix" and other kinds of non-tcp
socket types.  This patch preserves current behavior
of using "tcp" if not specified in the config, but
extends the config to allow specifying the gRPC
socket type.

### DIFF
--- a/rpc/grpc/config.go
+++ b/rpc/grpc/config.go
@@ -29,6 +29,10 @@ type Config struct {
 	// Endpoint is an address of GRPC netListener
 	Endpoint string `json:"endpoint"`
 
+	// Type of Socket defaults to "tcp" if unset
+
+	SocketType string `json:"socket-type"`
+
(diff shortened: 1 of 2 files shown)


jmedved:
Pls lets discuss the approaches in #273 and #274 and come up with a unified approach/API.

rpc/grpc/config.go
```diff
@@ -29,6 +29,10 @@ type Config struct {
 	// Endpoint is an address of GRPC netListener
 	Endpoint string `json:"endpoint"`
 
+	// Type of Socket defaults to "tcp" if unset
```
ondrej-fabry: You should also mention that `It must be "tcp", "tcp4", "tcp6", "unix" or "unixpacket".`, just like in [net.Listen]https://golang.org/pkg/net/#Listen).

PR #274, opened 2018-05-25
grpc unix domain socket support

VladoLavor:

### DIFF
--- a/rpc/grpc/config.go
+++ b/rpc/grpc/config.go
@@ -26,8 +26,11 @@ import (
 // Config is a configuration for GRPC netListener
 // It is meant to be extended with security (TLS...)
 type Config struct {
-	// Endpoint is an address of GRPC netListener
+	// Endpoint is an address of GRPC netListener for TCP/IP transport. Always chosen as default if set.
 	Endpoint string `json:"endpoint"`
+
+	// UnixSocketFilePath defines a path for unix socket file.
+	UnixSocketFilePath string `json:"unix-socket-file-path"`
 
 	// MaxMsgSize returns a ServerOption to set the max message size in bytes for inbound mesages.
 	// If this is not set, gRPC uses the default 4MB.
--- a/rpc/grpc/listen_and_serve.go
+++ b/rpc/grpc/listen_and_serve.go
@@ -15,9 +15,12 @@
 package grpc
 
 import (
+	"fmt"
 	"io"
 	"net"
 	"time"
+	"syscall"
+	"github.com/ligato/cn-infra/logging"
 
 	"google.golang.org/grpc"
 )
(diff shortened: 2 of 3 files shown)


rpc/grpc/config.go
```diff
@@ -26,9 +26,12 @@ import (
 // Config is a configuration for GRPC netListener
 // It is meant to be extended with security (TLS...)
 type Config struct {
-	// Endpoint is an address of GRPC netListener
+	// Endpoint is an address of GRPC netListener for TCP/IP transport. Always chosen as default if set.
 	Endpoint string `json:"endpoint"`
 
+	// UnixSocketFilePath defines a path for unix socket file.
+	UnixSocketFilePath string `json:"unix-socket-file-path"`
```
edwarnicke: Recommend we mirror how net.Listen works and simply specify the socketType (see https://github.com/ligato/cn-infra/pull/273 ).  What is the correct behavior if someone specifies both Endpoint and UnixSocketFilePath?

rpc/grpc/listen_and_serve.go
```diff
@@ -34,8 +37,20 @@ func FromExistingServer(listenAndServe ListenAndServe) *Plugin {
...
+func ListenAndServeGRPC(config *Config, grpcServer *grpc.Server, log logging.Logger) (netListener net.Listener, err error) {
+	// Start listener with TCP/IP or unix domain socket type
+	if config.Endpoint != "" {
+		netListener, err = net.Listen("tcp", config.Endpoint)
+		log.Info("Listening GRPC on tcp://", config.Endpoint)
+	} else if config.UnixSocketFilePath != "" {
+		// Remove old socket
+		syscall.Unlink(config.UnixSocketFilePath)
+		// Socket file is created automatically
+		netListener, err = net.Listen("unix", config.UnixSocketFilePath)
```
edwarnicke: What if we don't own the socket file (which will not always be the case).  Removing it is kind of aggressive.

rpc/grpc/plugin_impl_grpc.go
```diff
@@ -82,8 +82,7 @@ func (plugin *Plugin) Init() error {
 	if plugin.listenAndServe != nil {
 		plugin.netListener, err = plugin.listenAndServe(*plugin.grpcCfg, plugin.grpcServer)
 	} else {
-		plugin.Log.Info("Listening GRPC on tcp://", plugin.grpcCfg.Endpoint)
-		plugin.netListener, err = ListenAndServeGRPC(plugin.grpcCfg, plugin.grpcServer)
+		plugin.netListener, err = ListenAndServeGRPC(plugin.grpcCfg, plugin.grpcServer, plugin.Log)
```
edwarnicke: I don't understand how this relates to the unix domain socket.  This patch looks like two things (explicit adding of Logger to listen and serve and unix file socket got mushed together).

edwarnicke: Also, why in the default case do we want to force folks to manually pass round Loggers?

jmedved:
Ple let's discuss the relative merits of #273 and #274

edwarnicke:
I originally started where #274 is here (set a unix file name), but looking more closely at:
https://golang.org/pkg/net/#Listen

I realized that there were more network types... and so figured the most flexible forward compatible way to handle it was what was done in #273 with adding a field for the socketType (though one could easily argue I should have called it 'network' to match the naming in the net.Listen calll).

ondrej-fabry:
I must agree with @edwarnicke 's points here. 

The implementation in this PR just adds new config only for unix socket while being able to set old one to TCP address simultaneously allowing confusing configurations and restricting to just TCP or UNIX file socket while not preparing for some future network type without changing the config schema again. Also passing in loggers everywhere seems like a big anti-pattern, you either return an error or don't log anything.

I propose to revert this PR in favor of #273

PR #275, opened 2018-05-27
Introducing Wiring

edwarnicke:
Wiring packages with Flavors is hard and confusing.  It also encourages
overinclusion, because plugins come in big bundles some of which you
need and some of which you don't.

Wiring is a simple concept.  A plugin is Wireable if you can call:

`plugin.Wire(wiring)`

Wiring is just a function:

`type Wiring func (core.Plugin) error`

that is expected to wire up the plugins dependencies.

A plugin is said to be DefaultWireable if it implements:

`func DefaultWiring() Wiring`

so you can call

`plugin.Wire(plugin.DefaultWiring())`

conventionally, this is equivalent to `plugin.Wire(nil)`

Helper functions are provided to let you quickly get from a Wireable Plugin
to an Agent or running EventLoop

`agent, err :=wiring.NewAgent(plugin)`

`err := wiring.EventLoop(plugin)`

The net result of all of this is you can in one or two lines have a fully
functioning EventLoop running for your plugin.

This PR also reworks GRPC as a Wireable Plugin, introduces GRPC testing, and an example/wiring.


rpc/rest/wiring_impl.go
```diff
@@ -0,0 +1,161 @@
...
+		p, ok := plugin.(*Plugin)
+		if ok {
+			if len(cfg) > 0 {
+				p.PluginConfig = cfg[0]
+			} else {
+				if err := p.Wire(WithName(false)); err != nil {
+					return err
+				}
+				if overwrite || p.PluginConfig == nil {
+					p.PluginConfig = config.ForPlugin(p.Name())
```
ondrej-fabry: Missing extra parameter here makes example fail with error that the flag was not defined. Although it shouldn't be needed because it seems it's just a bug. 
See here: https://github.com/ligato/cn-infra/blob/master/config/plugin_config.go#L70
The if condition there fails because `len(opts)` is zero and that prevents registering the flag.

wiring/wiring_spi.go
```diff
@@ -0,0 +1,46 @@
...
+import "github.com/ligato/cn-infra/core"
+
+// Wiring is a function that, when called on a core.Plugin is expected to 'wire'
+// Its dependencies
+type Wiring func (plugin core.Plugin) error
+
+// Wireable implements a Wire function to wire in dependencies using a Wiring
+type Wireable interface {
+	Wire(wiring Wiring) error  // nil Wiring should result in a Default Wiring
+	IsWired() bool // true if the plugin is Wired, false otherwise
```
ondrej-fabry: The `IsWired` seems kind of unnecessary, shouldn't `DefaultWiring` take care of the proper initialization if not set otherwise?

ondrej-fabry:
The first problem I see with this approach is that now all of the plugins need to implement the `Wireable` interface and basically know about the wiring package at all. Compared to flavor approach, which is really controversial, but at least the only package that needs to import it is `main` package. That way it doesn't pollute all the code base and is quite separated. 

Also the `Wiring` function takes `core.Plugin` parameter which needs to be casted into proper `Plugin` type inside. Counting only your example where you implemented the wiring method for `rest`/`grpc` plugins you had to cast it 10 times! and those plugins got just basic dependencies! The vpp flavor we use in vpp-agent has over 20 plugins! and some of them have more than 5 dependencies! Can you imagine how many times would would have to be that boilerplate code implemented? And many many of them would have very similar `With..` options, just with different casting inside.

edwarnicke:
@ondrej-fabry 

What I am functionally trying to achieve is this:

A developer needs to be able to pick up a plugin, and just use it.

Today, if a plugin has N dependencies, the developer has to either:

a)  Hand wire all N dependencies that he neither understands nor wants to understand

or

b)  Find a Flavor that wires it for him.  But the Flavor also brings M *other* plugins he neither cares about nor wants

Developer walks away to find something more fun to do.

What Wirings allows the programmer to do is pick up the plugin and have it *just* work out of the box.

As the developer keeps working, maybe he figures out he wants to change *one* of the N dependencies.   Should be able to do that in the simplest possible way with the least possible understanding and work.

Note all plugins don't have to implement Wirable.  It's purely optional.  It's also purely orthogonal.  To prove that point I put all of the wiring implementations for grpc and its Wirings into a separate wiring.go file.  The code is also heavily patterned.  So heavily patterned that if you do what I did, and handle it modularly  with one Wiring per Dependency that you compose for things like your DefaultWiring, you wind up in situations like what I did where you can literally copy and paste between two plugins for a dependency they share in common both for the Wiring implementation and for the Wiring tests.

Note: I'm not using casts.  I'm using type assertions.  

Apparently, their performance has been discussed in detail:
https://stackoverflow.com/questions/28024884/does-a-type-assertion-type-switch-have-bad-performance-is-slow-in-go

And worst case looks like about 40ns per type assertion (read all the way down the thread).  So 10 type asserts * 40ns would be 400 ns.  400ns times 20 plugins would be 8000ns... which is 8 microseconds or 0.008 ms.  I like performance as much as the next guy, but that's an easy trade for me for usability :)

ondrej-fabry:
Sorry for my terminology, of course I meant **type assertion**, since there is no **type casting** in Go.

Also I have not mentioned the amount of type assertions due performance, but because of the amount of very similar code in many of the plugins for this to work with in larger projects. Performance isn't really a drive here since all of this just happens once during initialization.

I understand what your end goal is, I'm just trying to find a solution which will make it better for new devs with single plugins and also the large projects with tons of plugins and dependencies.

edwarnicke:
@ondrej-fabry I'm all for finding an approach that makes things simpler for new plugin developers.

I think it's important though to consider two classes there:

People doing simple extensions/aggregations of plugins (as I do in the example for greetingservice.Plugin and those building entirely new plugins.

For the later group, I think the Embedding trick takes most of this off their hands.  For example I had to write zero wiring code for greetingservice.Plugin.

I'm open to other suggestions that meeting the objectives :)

PR #280, opened 2018-05-29
Change SocketType->Network and improve comments

edwarnicke:
The eventual consumer of the value we are calling
SocketType is net.Listen or net.Dial
both of which call that value 'network'

Therefore it seemed wise to rename to align with them


rpc/grpc/config.go
```diff
@@ -29,9 +29,9 @@ type Config struct {
 	// Endpoint is an address of GRPC netListener
 	Endpoint string `json:"endpoint"`
 
-	// SocketType defaults to "tcp" if unset, and can be set to one of the following values:
-	// "tcp", "tcp4", "tcp6", "unix", "unixpacket"
-	SocketType string `json:"socket-type"`
+	// Network defaults to "tcp" if unset, and can be set to one of the following values:
+	// "tcp", "tcp4", "tcp6", "unix", "unixpacket" or any other value accepted by net.Listen
+	Network string `json:"socket-type"`
```
TrueFurby: Missed the struct tag.

edwarnicke: Good catch, should be fixed in next push :)

edwarnicke:
@TrueFurby I think I have now fixed your concerns about fixing the name in the tag as well.

PR #283, opened 2018-05-29
Provide an async Run method

edwarnicke:
readyCh,errCh = Run(agent,closeCh)

readyCh is closed when the agent it ready
- warning, if Agent.Start() returns an error
  readyCh is not closed
errCh receives errors and is closed as part of the
eventloop finishing

If there is an error starting the agent the err is
sent on the errCh and all channels are closed

errCh receives all errors

EventLoopWithInterupt was rewritten to use Run

### DIFF
--- a/core/event_loop.go
+++ b/core/event_loop.go
@@ -23,25 +23,42 @@ import (
 // EventLoopWithInterrupt starts an instance of the agent created with NewAgent().
 // Agent is stopped when <closeChan> is closed, a user interrupt (SIGINT), or a
 // terminate signal (SIGTERM) is received.
-func EventLoopWithInterrupt(agent *Agent, closeChan chan struct{}) error {
-	err := agent.Start()
-	if err != nil {
-		agent.Error("Error loading core: ", err)
-		return err
-	}
+func EventLoopWithInterrupt(agent *Agent, closeCh chan struct{}) error {
+	_, errCh := Run(agent, closeCh)
+	return <-errCh
+}
 
-	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, os.Interrupt)
-	signal.Notify(sigChan, syscall.SIGTERM)
-	select {
-	case <-sigChan:
-		agent.Println("Interrupt received, returning.")
-	case <-closeChan:
-	}
-
-	err = agent.Stop()
-	if err != nil {
-		agent.Errorf("Agent stop error '%+v'", err)
-	}
-	return err
+// Run starts an instance of the agent created with NewAgent().
+// Agent is stopped when <closeChan> is closed, , a user interrupt (SIGINT), or a
+// terminate signal (SIGTERM) is received.
+//
+// Returns readyCh which will be closed when the agent is ready - warning readyCh may never be closed if the agent fails on start
+// Returns errCh which will receive all errors associated with starting or stopping the event loop
+//    errCh will be closed when the agent is successfully stopped
+func Run(agent *Agent, closeCh chan struct{}) (<-chan struct{}, <-chan error) {
+	// We need two slots in the errCh, one for errors on Start(), one for errors on Stop()
+	errCh := make(chan error, 2)
+	readyCh := make(chan struct{})
+	go func() {
+		defer close(errCh)
+		if err := agent.Start(); err != nil {
+			agent.Error("Error loading core: ", err)
+			errCh <- err
+			return
+		}
+		close(readyCh)
+		sigChan := make(chan os.Signal, 1)
+		signal.Notify(sigChan, os.Interrupt)
+		signal.Notify(sigChan, syscall.SIGTERM)
+		select {
+		case <-sigChan:
+			agent.Println("Interrupt received, returning.")
+		case <-closeCh:
+		}
+		if err := agent.Stop(); err != nil {
+			agent.Errorf("Agent stop error '%+v'", err)
+			errCh <- err
+		}
+	}()
+	return readyCh, errCh
 }


core/event_loop.go
```diff
@@ -23,25 +23,46 @@ import (
...
 // Agent is stopped when <closeChan> is closed, a user interrupt (SIGINT), or a
 // terminate signal (SIGTERM) is received.
-func EventLoopWithInterrupt(agent *Agent, closeChan chan struct{}) error {
-	err := agent.Start()
-	if err != nil {
-		agent.Error("Error loading core: ", err)
-		return err
-	}
+func EventLoopWithInterrupt(agent *Agent, closeCh chan struct{}) error {
+	_,errCh := EventLoop(agent,closeCh)
```
ondrej-fabry: **gofmt!** 💥

core/event_loop.go
```diff
@@ -23,25 +23,46 @@ import (
...
-		agent.Error("Error loading core: ", err)
-		return err
-	}
+func EventLoopWithInterrupt(agent *Agent, closeCh chan struct{}) error {
+	_, errCh := EventLoop(agent, closeCh)
 
-	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, os.Interrupt)
-	signal.Notify(sigChan, syscall.SIGTERM)
 	select {
```
ondrej-fabry: `select` is unnecessary here when using single case. You could just do: `return <-errCh`.

core/event_loop.go
```diff
@@ -23,25 +23,46 @@ import (
...
+// Agent is stopped when <closeChan> is closed, , a user interrupt (SIGINT), or a
+// terminate signal (SIGTERM) is received.
+//
+// Returns readyCh which will be closed when the agent is ready - warning readyCh may never be closed if the agent fails on start
+// Returns errCh which will receive all errors associated with starting or stopping the event loop
+//    errCh will be closed when the agent is successfully stopped
+func EventLoop(agent *Agent, closeCh chan struct{}) (readyCh chan interface{}, errCh chan error) {
+	// We need two slots in the errCh, one for errors on Start(), one for errors on Stop()
+	errCh = make(chan error)
+	readyCh = make(chan interface{})
```
ondrej-fabry: You should generally use `chan struct{}` when not needing to send any data.

core/event_loop.go
edwarnicke: Good catch.  Fixed pushed :)

core/event_loop.go
```diff
@@ -23,25 +23,46 @@ import (
...
+// EventLoop starts an instance of the agent created with NewAgent().
+// Agent is stopped when <closeChan> is closed, , a user interrupt (SIGINT), or a
+// terminate signal (SIGTERM) is received.
+//
+// Returns readyCh which will be closed when the agent is ready - warning readyCh may never be closed if the agent fails on start
+// Returns errCh which will receive all errors associated with starting or stopping the event loop
+//    errCh will be closed when the agent is successfully stopped
+func EventLoop(agent *Agent, closeCh chan struct{}) (readyCh chan interface{}, errCh chan error) {
+	// We need two slots in the errCh, one for errors on Start(), one for errors on Stop()
+	errCh = make(chan error)
```
ondrej-fabry: This should be buffered channel otherwise it will block goroutine on send when no receiver is ready.

ondrej-fabry:
Well, to be honest I personally despised this whole `EventLoopWithInterrupt` function from the first time I saw it. Here's why:

- it actually doesn't really have anything to do with any [event loop](https://en.wikipedia.org/wiki/Event_loop), so the name is just confusing
- promotes indirection of the simplest calls: `Start`/`Stop` for the agent instance, which is actually needed only once in main so it doesn't really save lines or anything
- the only useful thing it really does is the waiting until signal is received

This enhancement proposed here kind of proves my point further.
I suppose that when using this function you realized that there is a need to know when agent started before it blocks waiting for signal (to eliminate sleeps in tests for example). And while this implementation might solve that case, it also brings a lot of weight with it as well. Specifically 2 extra channels and a goroutine, which certainly is insignificant for performance, but unnecessarily more complex to read for developers. 

What I propose is to get rid of this function entirely and move these essential agent calls into `main` function and perhaps just extract the signal waiting piece into separate function. 

**Before:**
```go
func main() {
	agent := flavor.NewAgent()
	err := core.EventLoopWithInterrupt(agent, nil)
	if err != nil {
		os.Exit(1)
	}
}
```

**After:**
```go
func main() {
	agent := flavor.NewAgent()

	if err := agent.Start(); err != nil {
		os.Exit(1)
	}

	agent.WaitForSignal(syscall.SIGINT, syscall.SIGTERM)

	if err := agent.Stop(); err != nil {
		os.Exit(1)
	}
}
```

core/event_loop.go
edwarnicke: Fixed.  Pushed.

edwarnicke:
@ondrej-fabry I see your point.

Do you see agent.WaitForSignal() as blocking?


Please also note I'm trying to get to:
```go
func main() {
      plugin := &foo.Plugin{}
      readyCh,errCh = plugin.EventLoop()
      doOtherStuff(readyCh, errCh)
} 
```

Lowest possible *developer* overhead using this and Wiring.  How would you do something this simple given your thoughts on doing this?

ondrej-fabry:
Yes, the `WaitForSignal` was meant as blocking call. It is just to extract the part with handling signals.

In your example what is that other stuff supposed to be? The `main` function should just wait there until signal. The bussiness logic should be inside plugins, that is what they are meant to be used for.

edwarnicke:
@ondrej-fabry doOtherStuff does whatever the developer wants it to.  It could be anything.  *Forcing* everything to be done through Ligato whether it makes sense or not (mandatory kidnapping of all of main) is a barrier to adoption.

The interesting question is if I am doing other stuff and a Signal comes in, does the Agent stop in your WaitForSignal model?

edwarnicke:
errCh is closed when Stopping wether there's an error or not, so you can always tell

```go
err, ok : <- errCh
```
whether the EventLoop has finished or not.  The developer can always manually manage these things as you've shown in your snippet.  The thing I am looking for is a way for the developer to offload having to do so.

Basically:

plugin.${callsomething} 

that does a good default right set of things in a way that you can tell when its ready, when its broken, and when it's done.

I can buy that there may be a better name for ${callsomething} ... do you have suggestions?

core/event_loop.go
```diff
@@ -23,25 +23,42 @@ import (
...
+// Agent is stopped when <closeChan> is closed, , a user interrupt (SIGINT), or a
+// terminate signal (SIGTERM) is received.
+//
+// Returns readyCh which will be closed when the agent is ready - warning readyCh may never be closed if the agent fails on start
+// Returns errCh which will receive all errors associated with starting or stopping the event loop
+//    errCh will be closed when the agent is successfully stopped
+func EventLoop(agent *Agent, closeCh chan struct{}) (readyCh chan struct{}, errCh chan error) {
+	// We need two slots in the errCh, one for errors on Start(), one for errors on Stop()
+	errCh = make(chan error)
+	readyCh = make(chan struct{}, 1)
```
ondrej-fabry: The `readyCh` channel doesn't need to be buffered, since no data is ever sent to it. Actually the `errCh` channel should be buffered, to prevent blocking of goroutine below since you send an error into it and there might not be any receiver.

core/event_loop.go
```diff
@@ -23,25 +23,42 @@ import (
...
-	}
-	return err
+// EventLoop starts an instance of the agent created with NewAgent().
+// Agent is stopped when <closeChan> is closed, , a user interrupt (SIGINT), or a
+// terminate signal (SIGTERM) is received.
+//
+// Returns readyCh which will be closed when the agent is ready - warning readyCh may never be closed if the agent fails on start
+// Returns errCh which will receive all errors associated with starting or stopping the event loop
+//    errCh will be closed when the agent is successfully stopped
+func EventLoop(agent *Agent, closeCh chan struct{}) (readyCh chan struct{}, errCh chan error) {
```
ondrej-fabry: You could rather return read-only channels since the caller of `EventLoop` is not supposed to send anything to `readyCh` and `errCh`.

ondrej-fabry:
@edwarnicke Okay, I guess I personally would suggest something like `Run` instead of `EventLoop`, which describes the behavior of it better, since it doesn't run any event loop, but *runs* agent: **Starts** -> **Waits for exit signal** -> **Stops**.

However I'd welcome some thoughts from others: @rastislavszabo @jmedved

core/event_loop.go
edwarnicke: Agreed.

core/event_loop.go
edwarnicke: Agreed.

edwarnicke:
@ondrej-fabry Honestly, I prefer something like run, but we have all of this existing use of EventLoop.  Given a naming convention I presume I can't change because its to entrenched, I went along with it.

ondrej-fabry:
We could just leave the `EventLoopWithInterrupt` for backwards compatibility and mention that it is deprecated.

edwarnicke:
@ondrej-fabry I like the suggestion of using Run and leaving EventLoopWithInterrupt for backward compatibility :)

core/event_loop.go
edwarnicke: Done

core/event_loop.go
edwarnicke: Done

edwarnicke:
Renamed EventLoop -> Run per suggestion

core/event_loop.go
```diff
@@ -23,25 +23,42 @@ import (
...
+// Run starts an instance of the agent created with NewAgent().
+// Agent is stopped when <closeChan> is closed, , a user interrupt (SIGINT), or a
+// terminate signal (SIGTERM) is received.
+//
+// Returns readyCh which will be closed when the agent is ready - warning readyCh may never be closed if the agent fails on start
+// Returns errCh which will receive all errors associated with starting or stopping the event loop
+//    errCh will be closed when the agent is successfully stopped
+func Run(agent *Agent, closeCh chan struct{}) (<-chan struct{}, <-chan error) {
+	// We need two slots in the errCh, one for errors on Start(), one for errors on Stop()
+	errCh := make(chan error, 2)
```
ondrej-fabry: There isn't any situation that `errCh` would need capacity of **2**.

* If the **start of agent fails**, error is sent on `errCh` channel and goroutine returns, which then closes the channel in defer, so no more errors will be sent.
* If the **start is ok** (nothing sent to `errCh` channel) and **stop of agent fails**, error is sent on `errCh` channel and goroutine ends after that, which then also closes the channel in defer, so again no more errors will be sent.

ondrej-fabry:
Actually now that I think about it, I'm not sure `Run` is the correct name for this. 

If someone were to use this `Run` function and wouldn't really care about when is agent ready (after successful Start), his example would have to at least look something like this:

```go
func main() {
    agent := core.NewAgent(...)
    _, errCh := core.Run(agent)
    if err := <-errCh; err != nil {
        log.Fatal(err)
    }
}
```

And I realized that `Run` much more describes blocking function, just like it is in std package exec [cmd.Run](https://golang.org/pkg/os/exec/#Cmd.Run):

> Run starts the specified command and waits for it to complete.

That's why I'd probably propose naming this function `RunAsync` (`Start` would be confusing with agent's `Start` method). And creating `Run` function which would do the same thing as `EventLoopWithInterrupt` and basically replace in the future and we would keep `EventLoopWithInterrupt` just as deprecated function.

What do you think?

PR #287, opened 2018-06-05
Fix bug in config.ForPlugin and add tests

edwarnicke:
config.ForPlugin was failing to work properly
in the case of no opts...

This patch fixes it, and introduces tests to insure proper
operation


config/config_test.go
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2017 Cisco and/or its affiliates.
```
ondrej-fabry: It's 2018.

config/config_test.go
```diff
@@ -0,0 +1,53 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config_test
+
+import (
+	"testing"
+
+	"github.com/ligato/cn-infra/config"
+	"github.com/onsi/gomega"
```
ondrej-fabry: Use dot import for more readable tests. 
`Expect(pluginConfig).ShouldNot(BeNil())` vs. `gomega.Expect(pluginConfig).ShouldNot(gomega.BeNil())`

config/config_test.go
edwarnicke: Done

config/config_test.go
edwarnicke: Done

PR #290, opened 2018-06-11
Make Init(), AfterInit(), and Close() threadsafe and idempotent

edwarnicke:
This patch makes Init(), AfterInit(), and Close() threadsafe and idempotent.
It has one commit per plugin that does it.

This is achieved by using sync.Once

An instance of sync.Once has method Do(func()) which will be called no more than once.
After the first time Do is called, it blocks until the first call of Do has finished.  If the first call of Do is already finished, then Do simply returns.


edwarnicke:
I had to remove the commit for https://github.com/ligato/cn-infra/tree/master/datasync/kvdbsync because the method OfDifferentAgent

https://github.com/ligato/cn-infra/blob/master/datasync/kvdbsync/plugin_impl_dbsync.go#L49

makes a copy of the plugin, including the mutexes in the sync.Once members, and go vet will have none of that.

This can be fixed by fixing OfDifferentAgent to bind to a pointer to the Plugin and having it make a new plugin and populate its fields with proper copies, including making new sync.Once's... but then we also need patches to the places its used in contiv-vpp.  So  job for later.

db/keyval/consul/plugin.go
```diff
@@ -88,50 +93,54 @@ func ConfigToClient(cfg *Config) (*api.Config, error) {
...
-					if plugin.Resync != nil {
-						plugin.Resync.DoResync()
-						plugin.lastConnErr = nil
-					} else {
-						plugin.Log.Warn("Expected resync after Consul reconnect could not start beacuse of missing Resync plugin")
+	plugin.initOnce.Do(func() {
+		cfg, err := plugin.getConfig()
+		if err != nil || plugin.disabled {
+			return
+		}
```
theckman: You've got a pretty gnarly bug here that means this will _never_ return an error. When you do `err := ...` you're shadowing the `err` variable you've defined at the top of the function. Shadowing means you're creating a variable with the same name, that prevents you from directly accessing the original variable.

Because you're shadowing the `err` variable defined in the function header, it means you aren't mutating the outer `err` variable. So when you do a bare `return` you're *_always_* going to return `nil`. Do you want to always return `nil`, or the error from `plugin.getConfig()`?

There are probably at least 10 other examples of that type of shadowing in this PR alone, and it seems there are more lurking in the codebase / other PRs.

PR #291, opened 2018-06-12
Make kvdbsync Init() and AfterInit() idempotent

edwarnicke:
... and also threadsafe.  Used sync.Once to do so.

This is complicated because of the binding of
OfDifferentAgent to a non-pointer version of Plugin.

Because OfDifferentAgent copies the entire Plugin
struct, including its Once's.  Once's contain
mutexes.  go vet screams bloody murder (rightly)
when you copy mutexes.

So I bound OfDifferentAgent to a pointer to the
original Plugin, created a new Plugin, and manually
copied the Deps.


datasync/kvdbsync/plugin_impl_dbsync.go
```diff
@@ -42,20 +45,27 @@ type infraDeps interface {
...
-// Method intentionally copies instance of a plugin (assuming it has set all dependencies)
-// and sets microservice label.
-func (plugin /*intentionally without pointer receiver*/ Plugin) OfDifferentAgent(
+// This method creates a new instance of a plugin that is supposed to watch different agent DB.
+// Method copies the existing plugins Deps and sets microservice label.
+func (p *Plugin) OfDifferentAgent(
 	microserviceLabel string, infraDeps infraDeps) *Plugin {
-
+	// Create a new plugin and copy over all the fields from the old one
+	plugin := &Plugin{}
```
ondrej-fabry: You should generally try to initialize all fields in single statement if nothing prevents that. 

Imagine you are debugging with debugger and are on breakpoint somewhere, splitting initialization of the fields separately will actually have plugin variable in different state on each step, but when initializing all fields in single statetement it will get fully initialized on the first step.

Also less code 😝

datasync/kvdbsync/plugin_impl_dbsync.go
```diff
@@ -67,10 +77,12 @@ type Deps struct {
...
 // Init only initializes plugin.registry.
-func (plugin *Plugin) Init() error {
-	plugin.registry = syncbase.NewRegistry()
+func (plugin *Plugin) Init() (err error) {
+	plugin.initOnce.Do(func() {
+		plugin.registry = syncbase.NewRegistry()
+	})
 
-	return nil
+	return err
```
ondrej-fabry: This is just confusing, `err` will always be `nil` here.

edwarnicke: It’s actually a protective pattern in this case.  The return value in this case comes from to us via closure on the func() being called by once.Do(). If someone subsequently alters the code so that there is an error, not having to remember to change this second return makes it less likely things will be screwed up.

Protect against future human error when it’s cheap.  It’s philosophically similar to putting defer close next to where you open something

edwarnicke: It’s actually a protective pattern in this case.  The return value in this case comes from to us via closure on the func() being called by once.Do(). If someone subsequently alters the code so that there is an error, not having to remember to change this second return makes it less likely things will be screwed up.

Protect against future human error when it’s cheap.  It’s philosophically similar to putting defer close next to where you open something

bweston92: Personally it would make me have to re-read the code to see where I missed `err` being updated. I think in this case returning `nil` would be preferred.

theckman: I'd give a :+1: to `return nil` and a :-1: to `return err` if it's always going to be nil.

PR #292, opened 2018-06-12
Introduce new agent and plugin options

ondrej-fabry:
Signed-off-by: Ondrej Fabry <ofabry@cisco.com>


ondrej-fabry:
@edwarnicke here it is in some working state, but probably some stuff will be changed.. especially part with NamedPlugin.. will have to decide what to do with it. If we should just add Name() method to Plugin interface or something else..

agent/agent.go
```diff
@@ -0,0 +1,88 @@
...
+	"os/signal"
+	"syscall"
+
+	"github.com/ligato/cn-infra/core"
+	"github.com/ligato/cn-infra/logging/logrus"
+	"github.com/namsral/flag"
+)
+
+type Agent interface {
+	Run() error
```
edwarnicke: What if I want to Stop() an agent?

agent/options.go
```diff
@@ -0,0 +1,180 @@
...
+type Options struct {
+	Version        string
+	MaxStartupTime time.Duration
+
+	Plugins []core.PluginNamed
+}
+
+func newOptions(opts ...Option) Options {
+	opt := Options{
+		Version:        "dev",
```
edwarnicke: Lets put "dev" in a constant.

examples/new/simple/main.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/ligato/cn-infra/examples/new/greeter"
+	"github.com/ligato/cn-infra/logging/logrus"
+	"github.com/ligato/cn-infra/rpc/grpc"
+	"google.golang.org/grpc/examples/helloworld/helloworld"
+)
+
+func main() {
+	// Create agent
+	a := agent.NewAgent(
+		agent.Version("1.0.0"),
```
edwarnicke: This needs defaults badly.  I shouldn't have to specify version by default, or startup time by default... if those aren't present we should get default values for them.

edwarnicke:
@ondrej-fabry How about this for lifecycle of the new Agent:

Run()  - Which in analogy to https://golang.org/pkg/os/exec/#Cmd.Run waits until the Agent finishes
Start() - Which in analogy to https://golang.org/pkg/os/exec/#Cmd.Start returns as soon as the Agent is running
Wait() - Which in analogy to https://golang.org/pkg/os/exec/#Cmd.Wait will Wait() until an Agent finishes
Stop() - Which would stop an Agent

Note:
Run() is exactly the same as Start().Wait()

agent/options.go
ondrej-fabry: It's used in a single place and it's there just to not be empty when not set. Actually we might keep the old way of having global vars in `version.go` file which are set by linker (`-ldflags`). I put it there as an example mainly.

agent/agent.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/namsral/flag"
+
+	"github.com/ligato/cn-infra/core"
+	"github.com/ligato/cn-infra/logging"
+	"github.com/ligato/cn-infra/logging/logrus"
+	"github.com/ligato/cn-infra/utils/once"
+)
+
+// Agent implements startup & shutdown procedures.
+type Agent interface {
```
VladoLavor: all methods deserve comments

agent/agent_test.go
```diff
@@ -0,0 +1,374 @@
...
+	defaultPluginName             = "testplugin"
+)
+
+func TestEmptyAgent(t *testing.T) {
+	RegisterTestingT(t)
+
+	agent := agent.NewAgent()
+	Expect(agent).NotTo(BeNil())
+	Expect(agent.Options()).NotTo(BeNil())
+	Expect(agent.Options().Version).To(Equal(defaultVersion))
```
VladoLavor: There is no 'Version' or 'MaxStartupTime' in Options

PR #293, opened 2018-06-12
Introduce once.ReturnError

edwarnicke:
This patch introduces a once.ReturnError that
behaves exactly like Once, but its Do method
takes

func() error

instead of

func()

To ensure proper handling of the error return.


edwarnicke:
@theckman Would this be a better way to approach the problem you noted here:

https://github.com/ligato/cn-infra/pull/290

PR #295, opened 2018-06-12
Make grpc.Plugin.{Init,AfterInit,Close}() idempotent and threadsafe

edwarnicke:
This PR is built on #293, only the last commit is unique to this commit

### DIFF
--- a/core/once_with_error.go
+++ b/core/once_with_error.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Cisco and/or its affiliates.
+// Copyright (c) 2018 Cisco and/or its affiliates.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/core/once_with_error.go
+++ b/core/once_with_error.go
@@ -1,0 +1,29 @@
+// Copyright (c) 2017 Cisco and/or its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


ondrej-fabry:
I'm really not sure about this approach. Why do we need this? 

I think that "protecting the code" this way actually promotes incorrect use of the `Init`/`AfterInit`/`Close` methods of plugins and puts extra requirements that need to followed when writing new plugins, which complicates the overall architecture. Even if you really think this is needed due some special case, then all of this could be achieved from the new agent package which initializes/closes plugins by saving the instances of plugins to some map and marking them as initialized/closed. 

And I think that the agent package should be the only place that calls `Init`/`AfterInit`/`Close` of plugins otherwise we gonna end up with too many ways of how the plugins are handled.I think allowing more ways to do same thing, initialization/closing of plugins, just encourages developers to write too varied cod. This is actually unidiomatic to Go, which tries to do stuff in one correct way.

rastislavszabo:
Personally, I would also vote for simplicity and readability over making the code extra safe, but not easy to follow

edwarnicke:
@ondrej-fabry As to why do we need this.  I need to be able to dynamically create GRPC plugins after Agent start time to support NSM.  Because of that need, I need to be able to Init() those plugins independently of the Agent.  Because of that, I need GRPC Init to be self-contained.   Being self contained implies it calls the Init of its dependencies.  The only way to do that *safely* is to ensure Init, AfterInit, and Close are idempotent.

PR #297, opened 2018-06-13
Make kafka.Plugin.{Init,AfterInit,Close}() idempotent and threadsafe

edwarnicke:
This PR is built on #293, only the last commit is unique to this commit

### DIFF
--- a/messaging/kafka/plugin_impl_kafka.go
+++ b/messaging/kafka/plugin_impl_kafka.go
@@ -27,6 +27,7 @@ import (
 	"github.com/ligato/cn-infra/messaging/kafka/client"
 	"github.com/ligato/cn-infra/messaging/kafka/mux"
 	"github.com/ligato/cn-infra/utils/clienttls"
+	"github.com/ligato/cn-infra/utils/once"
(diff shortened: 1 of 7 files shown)


messaging/kafka/plugin_impl_kafka.go
```diff
@@ -60,6 +65,10 @@ func FromExistingMux(mux *mux.Multiplexer) *Plugin {
 
 // Init is called at plugin initialization.
 func (plugin *Plugin) Init() (err error) {
+	return plugin.initOnce.Do(plugin.init)
+}
+
+func (plugin *Plugin) init() (err error) {
```
VladoLavor: this method is always called once (now I mean only from one place), so I would put is as an anonymous function to Init(). Also applies for AfterInit, Close and same functions in other patches

edwarnicke: @VladoLavor I can see your point of view.  The reason I did it this way is that it significantly improves readability and reduces probabilities of mistakes due to closure with an anonymous function.

PR #298, opened 2018-06-13
Continuation of Ondrej's Agent and Options work

edwarnicke:
This continues the work in PR #292


agent/agent.go
```diff
@@ -0,0 +1,206 @@
...
+	return stopCh
+}
+
+// Error returns any error that occurred when the agent was Stopped
+func (a *agent) Error() error {
+	// a.Stop() returns whatever error occurred when stopping the agent
+	// This is because of stopOnce
+	// If you try to retrieve an error before the agent is started, you will get
+	// an error complaining the agent isn't started.
+	return a.Stop()
```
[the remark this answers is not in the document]
edwarnicke: Interestingly, I originally did this with a.Stop().

Here's the use case.  A user uses <-a.After() in a select to wait on the agent and several other things.  a.After() comes back, and we want to retrieve the error.  It seemed unintuitive to call a.Stop() just to retrieve the error (though it would work).

This led me to thinking of other situations where one might want to retrieve the error from a Stopped agent, and so I added a.Error() to make it semantically unconfusing.

agent/options.go
```diff
@@ -0,0 +1,195 @@
...
+// Plugins creates an Option that adds a list of Plugins to the Agent's Plugin list
+func Plugins(plugins ...core.PluginNamed) Option {
+	return func(o *Options) {
+		o.Plugins = append(o.Plugins, plugins...)
+	}
+}
+
+// DescendantPlugins creates an Option that adds all of the Descendants of a Plugin Recursively to the Agent's
+// Plugin list
+func DescendantPlugins(plugin core.Plugin) Option {
```
ondrej-fabry: Hmm, perhaps `AllPlugins` might be simpler? And would actually mean that it will init all the plugins nested in this one.

edwarnicke: I can get behind AllPlugins 👍

agent/agent.go
```diff
@@ -0,0 +1,206 @@
...
+			select {
+			case <-sig:
+			case <-a.After():
+			}
+			logrus.DefaultLogger().Info("Signal received, stopping.")
+			// Doesn't hurt to call Stop twice, its idempotent because of the
+			// stopOnce
+			a.Stop()
+			signal.Stop(sig)
+			close(sig)
```
ondrej-fabry: Channels don't need to be closed after use. Garbage collector will take care of it.

agent/agent.go
```diff
@@ -0,0 +1,206 @@
...
+	// If we want to properly handle cleanup when a SIG comes in *during*
+	// agent startup (ie, clean up after its finished) we need to register
+	// for the signal before we start() the agent
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL)
+	err := a.start()
+	// If the agent started, we have things to clean up if here is a SIG
+	// So fire off a goroutine to do that
+	if err == nil {
+		go func() {
```
ondrej-fabry: Seems like this function should be extracted instead of `start`.

agent/agent.go
```diff
@@ -0,0 +1,206 @@
...
+			// Wait for signal or agent stop
+			select {
+			case <-sig:
+			case <-a.After():
```
ondrej-fabry: It is certain here that `a.shopCh` is create here, no need to go through exported API, you can use it directly.

edwarnicke: My experience has been that using internal details in multiple places is inherently more prone to error than using external APIs.   a.After() is tested.  a.stopCh is inherently not tested.

agent/agent.go
edwarnicke: start containers the actual logic for starting things.  Its thematically self contained.  This function thematically contains the logic for signal handling.    Putting both things in the same function becomes much harder to understand for subsequent readers.

agent/agent.go
edwarnicke: @ondrej-fabry True, but good hygiene is still good hygiene :)

ondrej-fabry: Unnecessary code = more code to read = longer time to understand. That line doesn't do anything useful, the signal watching is stopped, the goroutine ends and that channel will never be accessed again. Why keep useless code?

agent/agent.go
ondrej-fabry: My point is that `a.stopCh` doesn't even need to be tested, it's a **field**. And in this situation it is certain, since it's after successful start, that this field is there and it's just unnecessary to access this field indirectly.

agent/agent.go
```diff
@@ -0,0 +1,206 @@
...
+	return err
+
+}
+
+// Wait will not return until a SIGINT, SIGTERM, or SIGKILL is received
+// Or the Agent is Stopped
+// All Plugins are Closed() before Wait returns
+func (a *agent) Wait() error {
+	if a.stopCh != nil { // Don't wait if we didn't Start
+		<-a.After()
```
ondrej-fabry: By accessing the `stopCh` field through exported `After` function you are unnecessarily doing double check of `nil`, even though the outcome of `After` function is certain and will always be same. This just creates another layer of indirection = confusing readers.

agent/agent.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+		if postPlugin, ok := plug.(core.PostInit); ok {
+			if err := postPlugin.AfterInit(); err != nil {
+				return err
+			}
+		} else {
+			logrus.DefaultLogger().Debugf("plugin %v has no AfterInit", p)
+		}
+	}
+	a.stopCh = make(chan struct{}, 1) // If we are started, we have a stopCh to signal stopping
```
ondrej-fabry: Nothing is ever inserted into this channel, makes no sense making it buffered.

agent/agent.go
```diff
@@ -0,0 +1,206 @@
...
+
+// Options returns the Options the agent was created with
+func (a *agent) Options() Options {
+	return a.opts
+}
+
+// Start starts the agent.  Start will return as soon as the Agent is ready.  The Agent continues
+// running after Start returns.
+func (a *agent) Start() error {
+	return a.startOnce.Do(a.startSignalWrapper)
```
ondrej-fabry: I can't seem to see any logic behind this. "_Protecting the code_" this way just encourages incorrect use and adds extra layer of indirection. And more importantly this patter over `Start`/`Stop` prevents `agent` to be stopped and started again. So, this doesn't really make sense to allow multiple calls to start, but prevent to actually start again after stopping correctly.

You could simply do check if `stopCh` is `nil` and return error, since that's a way you are already using in other methods to check if agent was started or not. That is actually similar way how it's done in `os/exec` package in standard libary.

agent/agent.go
```diff
@@ -0,0 +1,206 @@
...
+func (a *agent) startSignalWrapper() error {
+	// If we want to properly handle cleanup when a SIG comes in *during*
+	// agent startup (ie, clean up after its finished) we need to register
+	// for the signal before we start() the agent
+	sig := make(chan os.Signal, 1)
+	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL)
+	err := a.start()
+	// If the agent started, we have things to clean up if here is a SIG
+	// So fire off a goroutine to do that
+	if err == nil {
```
ondrej-fabry: If we were to follow more idiomatic Go, you should first check if `err != nil` and return there instead of wrapping larger parts of normal code flow in if `err == nil` checks. Using this patter actually makes your function always end in happy path (`return nil`) and it's much easier to see various _checkpoints_ where it checks for errors.

agent/agent.go
edwarnicke: The fundamental question comes back to I think the following things:

1)  Do you want to recycle Agents (ie, be able to restart a stopped agent)
2) Are you willing to accept things going wrong if Start or Stop are called more than once?
3) Are you willing to accept things going wrong is Start and Stop are called while one or the other of them is running?
4) Are you willing to accept getting returned errors if start or stop are called more than once?
5) Are you willing to accept things going wrong if there is a race condition between two thread?

My answers would be no to all 5.  If you answer no to all five, then you can either:

a)  Use once.Do, which elegantly does 2-5.
b)  Protect stopCh with a mutex and use it as you suggest
c)  Introduce and mutex protect places to put startErr and stopErr
At which point you've reinvented once.Do, but in a much more confusing way.

Incorrect use is not something you can discourage, especially with potential race conditions.  It happens, and if you are lucky it blows up immediately.  If you are unlucky, it blows up unpredictably at random in the future.

agent/agent.go
edwarnicke: Agreed.

agent/agent.go
ondrej-fabry: 1. Depends.. someone might have use case for this, don't see technical limit of why it couldn't be allowed.

2. Yes, I would expect getting an error on second call to `Start` or `Stop`. Which would not hide the incorrect usage from devs and make it immediately obvious what they are doing wrong and make them fix their code instead of keeping pointless multiple calls to `Start`.
  
    This is similar way how they deal with this in **standard library**, calling `Start` for [exec.Cmd](https://godoc.org/os/exec#Cmd) while the process is already running will [return explicit error](https://golang.org/src/os/exec/exec.go#L354) that it was already started.

3. Yes, I expect things going wrong when I use some library *incorrectly*. Calling `Stop` while `Start` is running should not happen in *correct* code, meaning the normal usage dictates that we should call one after another or more specifically accessing agent only from one goroutine.

4. Explained in 2.

5. Definitely yes! I expect things going wrong when there is race condition! Race condition is a bug and should not be allowed to be kept in the code. You should be able to use tools to discover and fix them.
  
These kind of problems you are describing are also possible with `exec.Cmd` and I'm sure with ton of other things too. However they are not trying to "*protect the code*" from this. There aren't any `sync.Once` or mutexes in [(*Cmd).Start](https://golang.org/src/os/exec/exec.go#L338) method, which would allow calls from multiple goroutines. 

Sure, there are use cases for `sync.Once` even in the `os/exec` package, but they are mostly to prevent closing of files, specifically `stdin` pipe which might behave differently for different types of commands and not to allow accessing it from multiple goroutines.

By not "*protecting the code*" from these things you can easily **detect incorrect usage** of the program by running it with `-race` flag. With your pattern this would **not** be possible, because the `sync.Once` would hide this and would not uncover such incorrect code.

ondrej-fabry:
Maybe you could describe various ways you would expect to work with Agent.

edwarnicke:
At its simplest, I would expect users to Start() an agent, go off and do various other things, and perhaps at some point Stop() the agent.

The next run of complexity up from there is wanting to be able to *know* an Agent has started before you do something else.  In that case, you would want to call either Wait() or if you were waiting on several things After() (so you can select on it).

Because your program may not revolve around the Agent, I could easily see passing the agent to other parts of the code that might want to be *certain* the Agent has been Started, and so may call Start() themselves.  If you are not tightly coordinating across all of those places, your simplest route is to just call Start() again (thus the need for idem potence).

I guess the key point is: Agent as you are imagining it (and this is true of many things in Ligato) aren't very usable unless you have very tight coordination between all of the code interacting with them.  They demand omniscience.  Omniscience is *really* expensive.

ondrej-fabry:
> wanting to be able to know an Agent has **started** before you do something else. In that case, you would want to call either `Wait()`..

This will actually return only after agent is stopped in the current implementation.

> other parts of the code that might want to be certain the Agent has been Started..

This could be achieved with method like `IsStarted()` or `State()` and would actually give a way for the caller to know if it already was started before or it has to be started. 

With current implementation where having `Start()` is wrapped with `sync.Once` you have no way of knowing if it was started long ago or you just started it by calling `Start()` and also there isn't a way to just check without doing anything, where someone might execute certain action only if the agent is running and do nothing otherwise.

edwarnicke:
Yes, could be done with IsStarted() or State(), but then you need to still properly handle threadsafety and race conditions for those.  You could (and it may even be a smart idea) have a State() that indicates started or stopped.  But the complexity throughout the code in handling those in a threadsafe and race condition safe way would probably exceed using the once.Do to insure idempotence.

PR #299, opened 2018-06-13
Ligato CRD plug-in example

mestery:
This PR creates a new example around Custom Resource definitions.
This includes a full Ligato plugin example which will create
two different CRD resources.

Note: This commit pulls in a fair amount of vendor'd code, all of
which is required for CRDs.


mestery:
This should obviously be considered WIP, I don't have permission to label it that way however.

lukasmacko:
I found the code to be more "example-like" than a reusable plugin code that adds certain level of abstraction. Anyway, it's helpful to have it for those who plan to develop their own CRD. I'm just wondering whether it's worth introducing all k8s dependencies into cn-infra and subsequently into all dependant projects.

Since it's still work in progress I can't tell for sure but I think that it would be better to create a separate repo for it or keep it as a part of network service mesh project.

edwarnicke:
@lukasmacko The plugin may need a bit of generalization, but as cn-infra already drags in kafka, prometheus, and a host of others... I'm just not buying the "don't pull in k8s" thing :)

There's a *huge* need for a turnkey CRD solution, and if Ligato is willing to offer that, I could see it increasing use of Ligato by a lot.

mestery:
I'm fine with a separate Ligato repository for this @lukasmacko, but as Ed says, Ligato already pulls in so much. Either way, please provide guidance here. What is here is an example that obviously requires some work to integrate, but we could move this in the direction of providing a turnkey CRD solution, as @edwarnicke says.

lukasmacko:
In order to move this effort forward, let's try to identify a part of the code that can be reused during the development of another CRD. I'm not so familiar with the implementation details of CRD and I was not able to identify such a part at the first glance.

In other words let's say that I want to develop my own CRD @mestery, could you point me to a part I could leverage in order to ease the development or to avoid code duplication? Once we have that, it could be extracted/generalized as a cn-infra component/plugin. 
If we're not able to extract the general part, the current state of the pull request could be definitely moved to an example repository as-is.

mestery:
Closing this means you don't want this merged, which is fine, but did you want a separate repository so we can use it as an example? If so, let me know and I'll create a blank repository (ligato/crd-example) and push it there.

ondrej-fabry:
Closed after discussion with Jan. Here's your repo: https://github.com/ligato/crd-example

PR #300, opened 2018-06-14
GRPC improvements for unix-type network

VladoLavor:
* the plugin creates a directory for unix domain socket file if does not exist
* removes old socket file if exists
* ability to set permissions

### DIFF
--- a/rpc/grpc/config.go
+++ b/rpc/grpc/config.go
@@ -29,6 +29,9 @@ type Config struct {
 	// Endpoint is an address of GRPC netListener
 	Endpoint string `json:"endpoint"`
 
+	// Three or four-digit permission setup for unix domain socket file (if used)
+	Permission int `json:"permission"`
+
 	// Network defaults to "tcp" if unset, and can be set to one of the following values:
 	// "tcp", "tcp4", "tcp6", "unix", "unixpacket" or any other value accepted by net.Listen
 	Network string `json:"network"`
--- a/rpc/grpc/config.go
+++ b/rpc/grpc/config.go
@@ -32,6 +32,9 @@ type Config struct {
 	// Three or four-digit permission setup for unix domain socket file (if used)
 	Permission int `json:"permission"`
 
+	// If set and unix type network is used, the existing socket file will be always removed and re-created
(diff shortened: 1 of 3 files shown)


rpc/grpc/listen_and_serve.go
```diff
@@ -41,9 +47,31 @@ func ListenAndServeGRPC(config *Config, grpcServer *grpc.Server) (netListener ne
 	if socketType == "" {
 		socketType = "tcp"
 	}
-	netListener, err = net.Listen(socketType, config.Endpoint)
-	if err != nil {
-		return nil, err
+
+	if socketType == "unix" || socketType == "unixpacket" {
+		permissions, err := getUnixSocketFilePermissions(config.Permission)
+		if err != nil {
```
edwarnicke: We should have default permission that we use if the user doesn't specify.

edwarnicke: Ah wait, you do that further down :)

rpc/grpc/listen_and_serve.go
```diff
@@ -65,3 +93,42 @@ func ListenAndServeGRPC(config *Config, grpcServer *grpc.Server) (netListener ne
...
+		}
+		return os.FileMode(mode), nil
+	} else {
+		return os.ModePerm, nil
+	}
+}
+
+// Check old socket file/directory of the unix domain socket. Remove old socket file if exists or create the directory
+// path if does not exist.
+func checkUnixSocketFileAndDirectory(endpoint string) error {
```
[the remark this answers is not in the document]
VladoLavor: Well, contiv guys require it.  I'll make it configurable so everyone can be happy. By default, the existing socket file will NOT be removed, but you can force it in grpc.conf if you need.

rpc/grpc/config.go
```diff
@@ -32,6 +32,9 @@ type Config struct {
 	// Three or four-digit permission setup for unix domain socket file (if used)
 	Permission int `json:"permission"`
 
+	// If set and unix type network is used, the existing socket file will be always removed and re-created
```
edwarnicke: +1

PR #302, opened 2018-06-17
Implement Walk for plugins

edwarnicke:
This is not yet ready.

The intended final state is inspired by:
https://golang.org/pkg/path/filepath/#Walk

It basically consists of a function:
func Walk(core.Plugin,WalkFunc) error

Where
type Walkfunc func(core.Plugin) error

The idea would be that we could have WalkFunc like
List
Init
AfterInit
Close

That would walk down the plugin tree.  This would be instead
of the newly proposed options.Recursive.

Now, for what isn't right yet.
We have this really awesome pattern where we use Dep
to express our Dependencies and embed it in our Plugin.
Unfortunately, Dep is not a Plugin, so if you just walk down the trees
of things that are Plugins... that won't work.
So you actually have to walk the entire tree *looking* for Plugins

Note: the newly proposed options.Resurcive has exactly the same
problem.

I'm putting this out there just so folks can see what I'm thinking
even though its not done yet.

### DIFF
--- a/utils/pluginutils/walk_impl.go
+++ b/utils/pluginutils/walk_impl.go
@@ -1,0 +1,99 @@
+// Copyright (c) 2017 Cisco and/or its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


ondrej-fabry:
Why put this in separate package instead of keeping it in agent package? Agent package is all about managing plugins and splitting various functionality needed by agent just increases number of packages without any added value.

PR #307, opened 2018-07-12
allow delayed start for the etcd

VladoLavor:


db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -45,6 +49,10 @@ type Plugin struct {
 	autoCompactDone chan struct{}
 	reconnectResync bool
 	lastConnErr     error
+
+	// If plugin was not connected during init phase, the channel can be used to notify dbsync that the plugin was
+	// able to connect Redis after initialization, and receive notification that the registration was successful
```
lukasmacko: etcd instead of Redis

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -76,42 +85,45 @@ func (plugin *Plugin) Init() (err error) {
...
+						etcdCfg.Endpoints, err)
+				} else {
+					plugin.Log.Infof("ETCD server %s connected", etcdCfg.Endpoints)
+					// Configure connection and set as connected
+					plugin.configureConnection(etcdCfg)
+					plugin.connected = true
+					// Notify the dbsync that the database is available and can be registered
+					plugin.initNotifChan <- struct{}{}
+					// Wait until registration is done
+					<-plugin.initNotifChan
```
lukasmacko: There is a race condition in waiting for registration to be done. It would be better to use callback for the commands that needs to be executed on connect.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -76,42 +85,42 @@ func (plugin *Plugin) Init() (err error) {
...
-					if plugin.Resync != nil {
-						plugin.Resync.DoResync()
-						plugin.lastConnErr = nil
-					} else {
-						plugin.Log.Warn("Expected resync after ETCD reconnect could not start beacuse of missing Resync plugin")
-					}
+	if err != nil && etcdCfg.DelayedStart {
+		plugin.Log.Infof("ETCD server %s not reachable in init phase. Agent will continue to try to connect", etcdCfg.Endpoints)
+		// Even if the connection cannot be established during init, keep trying
+		go func(etcdCfg *Config) {
```
ondrej-fabry: This is too nested, it'd be better to extract it into separate function.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -30,13 +30,17 @@ import (
 const (
 	// healthCheckProbeKey is a key used to probe Etcd state
 	healthCheckProbeKey = "/probe-etcd-connection"
+	// ETCD reconnect interval
+	reconnectInterval = 2
```
ondrej-fabry: This should probably be configurable.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -76,42 +85,42 @@ func (plugin *Plugin) Init() (err error) {
...
+		plugin.Log.Infof("ETCD server %s not reachable in init phase. Agent will continue to try to connect", etcdCfg.Endpoints)
+		// Even if the connection cannot be established during init, keep trying
+		go func(etcdCfg *Config) {
+			for {
+				time.Sleep(reconnectInterval * time.Second)
+				plugin.Log.Infof("Connecting to ETCD %v ...", etcdCfg.Endpoints)
+				plugin.connection, err = NewEtcdConnectionWithBytes(*etcdClientCfg, plugin.Log)
+				if err != nil {
+					// Do not mark it as an error
+					plugin.Log.Debugf("Repeated attempt to connect to the ETCD %s failed. Err: %s",
```
ondrej-fabry: Do we really want to show failed retry as debug only?

It might not be obvious for someone looking at the log why it's not working, but at the same time we don't want to spam the log on longer period of unavailability.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -76,42 +85,42 @@ func (plugin *Plugin) Init() (err error) {
...
-				if plugin.reconnectResync && plugin.lastConnErr != nil {
-					plugin.Log.Info("Starting resync after ETCD reconnect")
-					if plugin.Resync != nil {
-						plugin.Resync.DoResync()
-						plugin.lastConnErr = nil
-					} else {
-						plugin.Log.Warn("Expected resync after ETCD reconnect could not start beacuse of missing Resync plugin")
-					}
+	if err != nil && etcdCfg.DelayedStart {
+		plugin.Log.Infof("ETCD server %s not reachable in init phase. Agent will continue to try to connect", etcdCfg.Endpoints)
```
ondrej-fabry: Mention here the time period after which the retry will be started.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -154,6 +183,43 @@ func (plugin *Plugin) Compact(rev ...int64) (toRev int64, err error) {
 	return 0, fmt.Errorf("connection is not established")
 }
 
+// If ETCD is connected, complete all other procedures
+func (plugin *Plugin) configureConnection(etcdCfg *Config) {
+	plugin.reconnectResync = etcdCfg.ReconnectResync
```
ondrej-fabry: I think it would be better to add `*Config` field to the plugin instead of just manually setting the fields.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -154,6 +183,43 @@ func (plugin *Plugin) Compact(rev ...int64) (toRev int64, err error) {
...
+		if etcdCfg.AutoCompact < time.Duration(time.Minute*60) {
+			plugin.Log.Warnf("Auto compact option for ETCD is set to less than 60 minutes!")
+		}
+		plugin.startPeriodicAutoCompact(etcdCfg.AutoCompact)
+	}
+	plugin.protoWrapper = kvproto.NewProtoWrapperWithSerializer(plugin.connection, &keyval.SerializerJSON{})
+
+	// Register for providing status reports (polling mode).
+	if plugin.StatusCheck != nil {
+		plugin.StatusCheck.Register(core.PluginName(plugin.PluginName), func() (statuscheck.PluginState, error) {
```
ondrej-fabry: This should definitely be registered even before the connection retries, so anyone reading the status can see *etcd* there with error. You can just check if `plugin.connection` is `nil` before actually trying to call `GetValue(...)`.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -131,12 +140,32 @@ func (plugin *Plugin) NewWatcher(keyPrefix string) keyval.ProtoWatcher {
 	return plugin.protoWrapper.NewWatcher(keyPrefix)
 }
 
+// DoResync performs ETCD resync
+func (plugin *Plugin) DoResync() {
+	plugin.Resync.DoResync()
```
ondrej-fabry: This will panic if `Resync` is not set.

db/keyval/etcd/config.go
```diff
@@ -40,6 +40,7 @@ type Config struct {
 	CAfile                string        `json:"ca-file"`
 	AutoCompact           time.Duration `json:"auto-compact"`
 	ReconnectResync       bool          `json:"resync-after-reconnect"`
+	DelayedStart          bool          `json:"delayed-start"`
```
ondrej-fabry: `DelayedRestart` sounds like we want the start to be delayed, even though the option actually means we want to allow delayed start/connect. 

I'd suggest to use either:
 - `DialRetry` which would also be more fitting with `DialTimeout` option we currently have
 - or at least put `AllowDelayedStart` to make it clear that we are allowing start to be delayed

@lukasmacko what do you think?

db/keyval/plugin_api_keyval.go
```diff
@@ -29,9 +31,20 @@ type KvProtoPlugin interface {
 	// The prefix is removed from the key retrieved by GetKey() in ProtoWatchResp.
 	// To avoid using a prefix, pass keyval.Root constant as argument.
 	NewWatcher(keyPrefix string) ProtoWatcher
+	// DoResync can be used to call full resync for key-value datastore
+	DoResync()
```
ondrej-fabry: Why would we need to define `DoResync()` as part of `KvProtoPlugin`?

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -131,12 +140,32 @@ func (plugin *Plugin) NewWatcher(keyPrefix string) keyval.ProtoWatcher {
...
 }
 
+// Connected returns *true* if the plugin has connection with the database.
+func (plugin *Plugin) Connected() bool {
+	return plugin.connected
+}
+
+// GetInitNotificationChan returns post-init notification channel
+func (plugin *Plugin) GetInitNotificationChan() <-chan func() {
+	return plugin.initNotifChan
```
ondrej-fabry: This will not work very well with multiple plugins that use etcd as dependency. Only one of them will be informed and the rest will block forever.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -76,42 +85,42 @@ func (plugin *Plugin) Init() (err error) {
...
+					plugin.Log.Debugf("Repeated attempt to connect to the ETCD %s failed. Err: %s",
+						etcdCfg.Endpoints, err)
+				} else {
+					plugin.Log.Infof("ETCD server %s connected", etcdCfg.Endpoints)
+					// Configure connection and set as connected
+					plugin.configureConnection(etcdCfg)
+					plugin.connected = true
+					resyncCallback := func() { plugin.DoResync() }
+					// Notify the dbsync that the database is available and can be registered
+					plugin.initNotifChan <- resyncCallback
```
ondrej-fabry: This will not work very well with multiple plugins that use etcd as dependency. Only one of them will be informed and the rest will block forever.

db/keyval/etcd/config.go
VladoLavor: I'm for AllowDelayedStart, that was the original name of the field, but it seemed too long for me

db/keyval/etcd/etcd.conf
```diff
@@ -35,4 +35,7 @@ resync-after-reconnect: false
 
 # Allow to start without connected ETCD database. Plugin will try to connect and if successful, overall resync will
 # be called
-delayed-start: true
\ No newline at end of file
+allow-delayed-start: true
+
+# Number of seconds between ETCD reconnect attempts. Has no use if `delayed start` is turned off
+reconnect-interval: 1
```
ondrej-fabry: Why show different value from the default one, here?

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -30,18 +30,30 @@ import (
...
+	// plugin config
+	config *Config
+
+	// List of callback functions, used in case ETCD is not connected immediately. All plugins using
+	// ETCD as dependency add their own function if cluster is not reachable. After connection, all
+	// functions are executed.
+	onConnection []func() error
+
 	autoCompactDone chan struct{}
 	reconnectResync bool
```
ondrej-fabry: We don't need this anymore. Just use value from `config` field.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -154,18 +163,99 @@ func (plugin *Plugin) Compact(rev ...int64) (toRev int64, err error) {
...
+		if plugin.config.AutoCompact < time.Duration(time.Minute*60) {
+			plugin.Log.Warnf("Auto compact option for ETCD is set to less than 60 minutes!")
+		}
+		plugin.startPeriodicAutoCompact(plugin.config.AutoCompact)
+	}
+	plugin.protoWrapper = kvproto.NewProtoWrapperWithSerializer(plugin.connection, &keyval.SerializerJSON{})
+}
+
+// Starts periodic ETCD status check
+func (plugin *Plugin) startStatusCheck() {
```
ondrej-fabry: This is way too nested. It will be much more readable to do the statuscheck registration in `Init` and use this function as `probe` argument for the `Register`.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -154,18 +163,99 @@ func (plugin *Plugin) Compact(rev ...int64) (toRev int64, err error) {
...
+				plugin.connected = false
+				return statuscheck.Error, fmt.Errorf("no ETCD connection available")
+			}
+			if _, _, _, err := plugin.connection.GetValue(healthCheckProbeKey); err != nil {
+				plugin.lastConnErr = err
+				plugin.connected = false
+				return statuscheck.Error, err
+			}
+			if plugin.reconnectResync && plugin.lastConnErr != nil {
+				plugin.Log.Info("Starting resync after ETCD reconnect")
```
ondrej-fabry: This log should probably be after the check if `Resync` is nil.

db/keyval/etcd/config.go
```diff
@@ -40,6 +40,8 @@ type Config struct {
 	CAfile                string        `json:"ca-file"`
 	AutoCompact           time.Duration `json:"auto-compact"`
 	ReconnectResync       bool          `json:"resync-after-reconnect"`
+	AllowDelayedStart     bool          `json:"allow-delayed-start"`
+	ReconnectInterval     int64         `json:"reconnect-interval"`
```
ondrej-fabry: All the other config options representing duration use `time.Duration`.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -154,18 +163,99 @@ func (plugin *Plugin) Compact(rev ...int64) (toRev int64, err error) {
...
+	}
+	plugin.Log.Infof("ETCD server %s not reachable in init phase. Agent will continue to try to connect every %d second(s)",
+		plugin.config.Endpoints, interval)
+	for {
+		time.Sleep(time.Duration(interval) * time.Second)
+		plugin.Log.Infof("Connecting to ETCD %v ...", plugin.config.Endpoints)
+		plugin.connection, err = NewEtcdConnectionWithBytes(*clientCfg, plugin.Log)
+		if err != nil {
+			continue
+		} else {
```
ondrej-fabry: The `else` case is not needed.

datasync/kvdbsync/plugin_impl_dbsync.go
```diff
@@ -80,7 +80,27 @@ func (plugin *Plugin) Init() error {
...
 // since Watch() is called in Plugin.Init() and Resync.Register()
 // is called in Plugin.AfterInit().
+//
+// If provided connection is not ready (not connected), AfterInit starts new goroutine in order to
+// 'wait' for the connection. After that, the new transport watcher is built as usual.
 func (plugin *Plugin) AfterInit() error {
+	if plugin.KvPlugin == nil || plugin.KvPlugin.Disabled() {
+		return nil
+	}
+	if plugin.KvPlugin.Connected() {
```
ondrej-fabry: I believe there is a race condition here. The `Connected` returns false and then before you call `OnConnect` to register the handler, it connects and the handler is never run.

One solution for this could be to check inside the `OnConnect` before adding handler to the slice if the connection is already established and run the function from there. Then the `Connected` method might actually be obsolete because you could just use `OnConnect`.

db/keyval/etcd/etcd.conf
VladoLavor: because it's config file example. Why set default values via the config file?

db/keyval/etcd/plugin_impl_etcd.go
VladoLavor: I think this log is useless

datasync/kvdbsync/plugin_impl_dbsync.go
VladoLavor: you mean callback, not handler

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -31,7 +31,7 @@ const (
 	// healthCheckProbeKey is a key used to probe Etcd state
 	healthCheckProbeKey = "/probe-etcd-connection"
 	// ETCD reconnect interval
-	defaultReconnectInterval = 2
+	defaultReconnectInterval time.Duration = 2
```
ondrej-fabry: This here means `defaultReconnectInterval` is `2` nanoseconds. You need to use `x * time.Second`.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -213,34 +217,28 @@ func (plugin *Plugin) configureConnection() {
...
-				plugin.Log.Info("Starting resync after ETCD reconnect")
-				if plugin.Resync != nil {
-					plugin.Resync.DoResync()
-					plugin.lastConnErr = nil
-				} else {
-					plugin.Log.Warn("Expected resync after ETCD reconnect could not start beacuse of missing Resync plugin")
-				}
+// ETCD status check probe function
+func (plugin *Plugin) statusCheckProbe() statuscheck.PluginStateProbe {
+	return func() (statuscheck.PluginState, error) {
```
ondrej-fabry: What's the reason to return function here? You can just define `statusCheckProbe` to have this signature and use `plugin.StatusCheck.Register(plugin.PluginName, plugin.statusCheckProbe)` instead of `plugin.StatusCheck.Register(plugin.PluginName, plugin.statusCheckProbe())` to register it.

db/keyval/etcd/etcd.conf
```diff
@@ -31,4 +31,11 @@ auto-compact: true
...
-resync-after-reconnect: false
\ No newline at end of file
+resync-after-reconnect: false
+
+# Allow to start without connected ETCD database. Plugin will try to connect and if successful, overall resync will
+# be called
+allow-delayed-start: true
+
+# Number of seconds between ETCD reconnect attempts. Has no use if `delayed start` is turned off
+reconnect-interval: 1
```
ondrej-fabry: This needs to be changed because our parses parses it like raw `time.Time` (`uint64`), meaning it's in nanoseconds.

datasync/kvdbsync/plugin_impl_dbsync.go
ondrej-fabry: Sure, callback handler 🤣

db/keyval/etcd/etcd.conf
ondrej-fabry: > Why set default values via the config file?

Precisely because of the reason you mentioned: *it's config example* and should mainly serve as documentation of parameters, but also as starting point for users to see what the default value is without going through the code or logs.

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -30,20 +30,31 @@ import (
 const (
 	// healthCheckProbeKey is a key used to probe Etcd state
 	healthCheckProbeKey = "/probe-etcd-connection"
+	// ETCD reconnect interval
+	defaultReconnectInterval time.Duration = 2000000000 // 2 seconds
```
ondrej-fabry: Please define this as `time.Second * 2`. 🙄

db/keyval/etcd/plugin_impl_etcd.go
```diff
@@ -31,7 +31,7 @@ const (
 	// healthCheckProbeKey is a key used to probe Etcd state
 	healthCheckProbeKey = "/probe-etcd-connection"
 	// ETCD reconnect interval
-	defaultReconnectInterval time.Duration = 2000000000 // 2 seconds
+	defaultReconnectInterval = 2 * time.Second // 2 seconds
```
ondrej-fabry: No need for comment to repeat the code.

PR #310, opened 2018-07-16
Add cryptodata plugin used to encrypt/decrypt data

deathbeam:
Add cryptodata plugin that will provide support for encrypting and
decrypting arbitrary data using configured private keys and support for
wrapping k/v BytesBroker to automatically decrypt any read value if it
matches specified pattern.

Signed-off-by: Tomas Slusny <slusnucky@gmail.com>


db/keyval/etcd/config.go
```diff
@@ -96,6 +102,13 @@ func ConfigToClient(yc *Config) (*ClientConfig, error) {
 		}
 	}
 
+	if yc.Privatekeyfile != "" {
+		privateKey, err := keyval.ReadPrivateKey(yc.Privatekeyfile)
+		if err == nil {
+			cfg.PrivateKey = privateKey
```
ondrej-fabry: This should fail it error probably, since we failed to read private key we configured.

db/keyval/cryptodata/bytes_broker.go
```diff
@@ -0,0 +1,120 @@
...
+// Then it parses data as JSON as tries to lookup all top-level values that begin with $crypto$ and decrypt them
+// using provided private key.
+func decryptJSONValue(inData []byte, privateKey *rsa.PrivateKey) (data []byte) {
+	data = inData
+
+	if privateKey == nil {
+		return
+	}
+
+	hasEncrypted := regexp.MustCompile(`"encrypted"\s*:\s*"true"`)
```
ondrej-fabry: This needs to run only once. Make it package level variable.

ondrej-fabry:
The main problem I see here is that now all the packages; etcd, consul and redis need to know about the extra setting in config and dependency on the cryptodata package. All future implementations along with all the possible implementations outside this package have to implement stuff to support it too.

We don't want that. The configuration should be at once place, in the cryptodata package, where it should probably become plugin as well. The etcd and other packages should not need to know that anything about the crypto. The cryptodata package should provide some API to turn anyone implementing the BytesBrokerWatcher...API to wrap it in some kind of *crypto layer* or *crypto middleware* or whatever you would call it, and it will do its magic without anyone else changing. 

Using this method you will provide this crypto data processing for all current and future implementations if keyval storages.

VladoLavor:
I think it makes sense to create a plugin from cryptodata. I would also move it away from keyval package to the cn-infra/db/crypto... (you can probably use it for cassandra as well). I also agree that database plugins should not know about the encryption. We can also think about whether we store private keys to respective plugin config, or create s kind of crypto config with all the private keys for all db plugins.

ondrej-fabry:
@deathbeam Ping me before you continue your work on this, we will go over the implementation details.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	jsonMap, ok := jsonData.(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	for k, v := range jsonMap {
+		stringVal, ok := v.(string)
+		prefix := "$crypto$"
```
ondrej-fabry: This seems like something that should be defined outside, perhaps as a constant.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,79 @@
...
+// Then it parses data as JSON as tries to lookup all top-level values that begin with $crypto$ and decrypt them
+// using provided arbitrary decrypt function.
+func (DecrypterJSON) Decrypt(inData []byte, decryptArbitrary DecryptArbitrary) (data []byte) {
+	data = inData
+
+	if !encryptedJSONRegex.Match(inData) {
+		return
+	}
+
+	var jsonData interface{}
```
ondrej-fabry: You could perhaps define this as `map[string]interface{}` to avoid converting later.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,79 @@
...
+		return
+	}
+
+	jsonMap, ok := jsonData.(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	for k, v := range jsonMap {
+		stringVal, ok := v.(string)
```
ondrej-fabry: What if the value is `[]byte`, which is much more common. You can use type switch when you check if the value is one of the several types.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// DecrypterJSON is Decrypter implementations that can decrypt JSON values
+type DecrypterJSON struct{}
+
+// Decrypt tries to decrypt JSON data that are encrypted based on "encrypted": "true" presence in it.
+// Then it parses data as JSON as tries to lookup all top-level values that begin with $crypto$ and decrypt them
+// using provided arbitrary decrypt function.
+func (DecrypterJSON) Decrypt(inData []byte, decryptArbitrary DecryptArbitrary) (data []byte) {
+	data = inData
```
ondrej-fabry: This does not make a copy of the slice if that's what you intended to do.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,79 @@
...
+
+var encryptedJSONRegex = regexp.MustCompile(`"encrypted"\s*:\s*"true"`)
+
+// DecryptArbitrary is function that decrypts arbitrary data
+type DecryptArbitrary func(inData []byte) (data [] byte, err error)
+
+// Decrypter is interface for decrypting groups of data
+type Decrypter interface {
+	// Decrypt decrypts input data using provided decrypting function for arbitrary data
+	Decrypt(inData []byte, decryptArbitrary DecryptArbitrary) (data []byte)
```
ondrej-fabry: The method `Decrypt` should return error as well, how else will the caller know if the decrypt succeeded?

db/cryptodata/plugin.go
```diff
@@ -0,0 +1,100 @@
...
+func (plugin *Plugin) Disabled() bool {
+	return plugin.disabled
+}
+
+// UpdateConfig updates private key configuration for plugin
+func (plugin *Plugin) UpdateConfig(config Config) (err error) {
+	var clientConfig ClientConfig
+	err = ReadCryptoConfig(config, &clientConfig)
+	if err == nil {
+		plugin.privateKeys = clientConfig.PrivateKeys
```
ondrej-fabry: Why not return `ClientConfig` from `ReadCryptoConfig` and also have it as a field in `Plugin` instead of defining the same field, `privateKeys` in both.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// DecrypterJSON is Decrypter implementations that can decrypt JSON values
+type DecrypterJSON struct{}
+
+// Decrypt tries to decrypt JSON data that are encrypted based on "encrypted": "true" presence in it.
+// Then it parses data as JSON as tries to lookup all top-level values that begin with $crypto$ and decrypt them
+// using provided arbitrary decrypt function.
+func (DecrypterJSON) Decrypt(inData []byte, decryptArbitrary DecryptArbitrary) (data []byte) {
```
ondrej-fabry: I think the naming is actually inverted here. 

The `Decrypt` should simply just decrypt data, essentially what `DecryptArbitrary` does now.
And the `DecryptArbitrary` should try to look fields marked as encrypted and try to decrypt them using `Decrypter`.
 
Actually instead of defining `Decrypter` interface here, you could use one already defined in standard libary: https://godoc.org/crypto#Decrypter

db/cryptodata/config.go
```diff
@@ -0,0 +1,69 @@
...
+
+// ClientConfig is result of converting Config.PrivateKeyFile to PrivateKey
+type ClientConfig struct {
+	// Private key is used to decrypt encrypted keys while reading them from store
+	PrivateKeys []*rsa.PrivateKey
+}
+
+// ReadCryptoConfig reads private key from PEM path
+func ReadCryptoConfig(config Config, clientConfig *ClientConfig) (error) {
+	if len(config.PrivateKeyFiles) == 0 {
```
ondrej-fabry: This is unnecessary, for loop won't be executed if the `PrivateKeyFiles` are empty or nil and the function will return nil either way.

db/cryptodata/config.go
```diff
@@ -0,0 +1,69 @@
...
+		return nil
+	}
+
+	for _, file := range config.PrivateKeyFiles {
+		bytes, err := ioutil.ReadFile(file)
+		if err != nil {
+			return err
+		}
+
+		block, _ := pem.Decode(bytes)
```
ondrej-fabry: Reading the docs for `pem.Decode` it seems to me that it can support multiple keys, by returning the rest of file after first key.

So, perhaps loop can be used here and after successful decode, when `p != nil`, it should try to decode the rest in next iteration until we get `p == nil`.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,79 @@
...
+	"encoding/json"
+	"strings"
+)
+
+const cryptoPrefix = "$crypto$"
+
+var encryptedJSONRegex = regexp.MustCompile(`"encrypted"\s*:\s*"true"`)
+
+// DecryptArbitrary is function that decrypts arbitrary data
+type DecryptArbitrary func(inData []byte) (data [] byte, err error)
```
ondrej-fabry: This should be called `DecryptFunc` to follow more idiomatic way for types that represent functions.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Decrypter is interface for decrypting groups of data
+type Decrypter interface {
```
ondrej-fabry: And this should actually be called something like `ArbitraryDecrypter`. Since it looks inside arbitrary data to find specific fields marked as encrypted.

db/cryptodata/client.go
```diff
@@ -0,0 +1,74 @@
...
+	// If reader is nil use default rand.Reader
+	if client.Reader == nil {
+		client.Reader = rand.Reader
+	}
+
+	return
+}
+
+// EncryptArbitrary decrypts arbitrary input data using provided public key
+func (client *Client) EncryptArbitrary(inData []byte, pub *rsa.PublicKey) (data []byte, err error) {
```
ondrej-fabry: Let's name this `EncryptData`, since it encrypts all input.

db/cryptodata/client.go
```diff
@@ -0,0 +1,74 @@
...
+	data, err = rsa.EncryptPKCS1v15(client.Reader, pub, inData)
+	return
+}
+
+// DecryptArbitrary decrypts arbitrary input data
+func (client *Client) DecryptArbitrary(inData []byte) (data []byte, err error) {
```
ondrej-fabry: .. and `DecryptData` here.

db/cryptodata/plugin.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			return err
+		}
+
+		block, rest := pem.Decode(bytes)
+		if block == nil {
+			return errors.New("failed to decode PEM for key " + file)
+		}
+
+		for {
```
ondrej-fabry: This for loop should be defined as `for block != nil {` otherwise the call on line 73 can crash by accessing nil pointer if the previous iteration failed to decode anything from rest.

db/cryptodata/plugin.go
```diff
@@ -0,0 +1,101 @@
...
+			privateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+			if err != nil {
+				return err
+			}
+
+			privateKey.Precompute()
```
ondrej-fabry: We should probably use `Precompute` after we check that the key is valid.

deathbeam:
Updated. Additionally had to base64 encode/decode the data in order for them to be valid string for JSON.

Example output from cryptodata-lib sample:
```
root@bb850be6bf78:/go/src/github.com/ligato/cn-infra/examples/cryptodata-lib# go build
root@bb850be6bf78:/go/src/github.com/ligato/cn-infra/examples/cryptodata-lib# ./cryptodata-lib hello-world
Input hello-world
Encrypted pLO4qaN5VD0ss9KDTLtdklNfwaHw2ZLrSRJQnlgqPp_eR8Up8gXowCYqw4dj1OdlKc6R-kUQGp6rA424h34CCx2zMqS8i9VA1_ATnRc1hVL9G2PVVlRl3uS36x0Bstn6gHYbW2_gcHoSXXrBg6keW78XO2tWKuhMKBVzubpsL8sOewg2Ja4q778BitovmakfNDRiPnzHPGna9Rxk7tXyLJnWqC5FBeI-yZD7UXeJ_S-MEi83YePchegbR2rwmF3w6VL3sxuj3Z2gMFv1yZrZKi0pCt5SOILxSfhZYIIMXF0kwG5--kF-5MgvgSj9RonR4i246ejt1DSfoH5fON9f4A==
Decrypted hello-world
Encrypted json
{ "encrypted": "true", "payload": "$crypto$pLO4qaN5VD0ss9KDTLtdklNfwaHw2ZLrSRJQnlgqPp_eR8Up8gXowCYqw4dj1OdlKc6R-kUQGp6rA424h34CCx2zMqS8i9VA1_ATnRc1hVL9G2PVVlRl3uS36x0Bstn6gHYbW2_gcHoSXXrBg6keW78XO2tWKuhMKBVzubpsL8sOewg2Ja4q778BitovmakfNDRiPnzHPGna9Rxk7tXyLJnWqC5FBeI-yZD7UXeJ_S-MEi83YePchegbR2rwmF3w6VL3sxuj3Z2gMFv1yZrZKi0pCt5SOILxSfhZYIIMXF0kwG5--kF-5MgvgSj9RonR4i246ejt1DSfoH5fON9f4A==" }
Decrypted json
{"encrypted":"true","payload":"hello-world"}
```

db/cryptodata/client.go
```diff
@@ -0,0 +1,81 @@
...
+		client.Reader = rand.Reader
+	}
+
+	return
+}
+
+// EncryptData encrypts input data using provided public key
+func (client *Client) EncryptData(inData []byte, pub *rsa.PublicKey) (data []byte, err error) {
+	data, err = rsa.EncryptPKCS1v15(client.Reader, pub, inData)
+	data = []byte(base64.URLEncoding.EncodeToString(data))
```
ondrej-fabry: Why would we need to encode the data using encoder for URLs?

db/cryptodata/client.go
```diff
@@ -0,0 +1,81 @@
...
+	if client.Reader == nil {
+		client.Reader = rand.Reader
+	}
+
+	return
+}
+
+// EncryptData encrypts input data using provided public key
+func (client *Client) EncryptData(inData []byte, pub *rsa.PublicKey) (data []byte, err error) {
+	data, err = rsa.EncryptPKCS1v15(client.Reader, pub, inData)
```
ondrej-fabry: Reading the docs for `EncryptPKCS1v15` it seems it's not very secure for using for anything other than session keys.

> WARNING: use of this function to encrypt plaintexts other than session keys is dangerous. Use RSA OAEP in new protocols. 
https://godoc.org/crypto/rsa#EncryptPKCS1v15

And this will definitely be used for something else than just session keys.

db/cryptodata/client.go
ondrej-fabry: Nvm, just read the comment.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,79 @@
...
+		return
+	}
+
+	for k, v := range jsonData {
+		var stringVal string
+		switch t := v.(type) {
+		case string:
+			stringVal = t
+		case []byte:
+			stringVal = string(t)
```
ondrej-fabry: You shouldn't convert from `[]byte` to `string` if it's not needed. 

Here you are actually doing it multiple times. Once here to `string` and then below back to `[]byte`, that is not very efficient. You can use `bytes` package for working with bytes.

Actually.. reading the docs for `json.Unmarshal` it seems the `[]byte` will never be used for any type of data in json. So scratch my previous suggestion to add it here.

However what might be an issue is nested data. Since you are only going over top level fields, this won't work for json data like this:

```json
{
  "encrypted": true,
  "nested": {
    "abc": "aaa",
    "zxy": "$crypto$xxxxxxxx"
  }
}
```

You should probably do some kind of loop that will check if the type is `map[string]interface{}` and if so then do another loop to check for encrypted fields.

db/cryptodata/client.go
```diff
@@ -0,0 +1,90 @@
...
+	if client.Hash == nil {
+		client.Hash = sha256.New()
+	}
+
+	return
+}
+
+// EncryptData encrypts input data using provided public key
+func (client *Client) EncryptData(inData []byte, pub *rsa.PublicKey) (data []byte, err error) {
+	data, err = rsa.EncryptOAEP(client.Hash, client.Reader, pub, inData, nil)
```
ondrej-fabry: You should check for error here and return immediately if it's not nil.

db/cryptodata/client.go
```diff
@@ -0,0 +1,90 @@
...
+	data = []byte(base64.URLEncoding.EncodeToString(data))
```
ondrej-fabry: Actually, this might be required in json. However it's gonna be unnecessary for types that work with `[]byte`, since it will unnecessarily make it larger. It seems to me it should be used only for the json decrypter.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+// Decrypt tries to decrypt JSON data that are encrypted based on "encrypted": "true" presence in it.
+// Then it parses data as JSON as tries to lookup all top-level values that begin with $crypto$ and decrypt them
+// using provided arbitrary decrypt function.
+func (d DecrypterJSON) Decrypt(inData []byte, decryptFunc DecryptFunc) (data []byte, err error) {
+	data = inData
+
+	if !d.Present.Match(inData) {
```
ondrej-fabry: It might actually be faster to try to unmarshal the json data into struct like this:

```go
type EncryptionCheck struct {
    IsEncrypted bool `json:"encrypted"`
}
```

I can't say for sure, would have to do some benchmarks. However the implementation of `DecrypterJSON` should allow using different method entirely instead of forcing use of regexp. The `Present` field should be something like a function which takes in data and return bool if the data is encrypted or not.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,101 @@
...
+	data, err = json.Marshal(jsonData)
+	return
+}
+
+// decryptJSON recursively navigates JSON structure and tries to decrypt all string values with Prefix
+func (d DecrypterJSON) decryptJSON(data map[string]interface{}, decryptFunc DecryptFunc) (map[string]interface{}, error) {
+	for k, v := range data {
+		switch t := v.(type) {
+		case string:
+			if !strings.HasPrefix(t, d.Prefix) {
```
ondrej-fabry: You could combine this with the trimming to save extra call, like this:

```go
if s := strings.TrimPrefix(t, d.Prefix); s != t {
// ...
```

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+
+			data[k] = string(arbitraryData)
+		case map[string]interface{}:
+			val, err := d.decryptJSON(t, decryptFunc)
+			if err != nil {
+				return nil, err
+			}
+
+			v = val
```
ondrej-fabry: What's this for? Isn't `v` gonna be overwritten in the next iteration?

examples/cryptodata-lib/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// Decrypt previously encrypted input string
+	decrypted, err := client.DecryptData(encrypted)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Printf("Decrypted %v\n", string(decrypted))
+
+	// Try to decrypt JSON with encrypted data
+	encryptedJSON := fmt.Sprintf(`{ "encrypted": "true", "payload": "$crypto$%v" }`, string(encrypted))
```
ondrej-fabry: This should also show an example with encrypted data being deeper than just root level.

ondrej-fabry:
Could you make another example where it'd be used as a plugin and show usage of the `Wrap` function.

It can be simple just putting some data to Etcd with encrypted part in it and then showing usage of wrapped `Get` to see how we receive decrypted output.

You can base it off https://github.com/ligato/cn-infra/tree/master/examples/datasync-plugin
Minus the parts with Watch/List/etc.

examples/cryptodata-plugin/main.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return err
+	}
+
+	decryptedJSON, _, _, err := dbWrapped.GetValue(key)
+	if err != nil {
+		return err
+	}
+
+	plugin.Log.Info("Got value %v", string(decryptedJSON))
```
ondrej-fabry: Should be `Infof`.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,113 @@
...
+	"encoding/json"
+	"strings"
+	"encoding/base64"
+)
+
+// DecryptFunc is function that decrypts arbitrary data
+type DecryptFunc func(inData []byte) (data [] byte, err error)
+
+// ValidateFunc is used to validate compatibility of JSON data when decrypting
+type ValidateFunc func(inData []byte) (isValid bool)
```
ondrej-fabry: This naming doesn't fit here. It doesn't validate the JSON data, it just checks if the data contains encrypted data so it can start looking for it.

Perhaps `CheckEncryptedFunc`?

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,113 @@
...
+	// Decrypt decrypts input data using provided decrypting function for arbitrary data
+	Decrypt(inData []byte, decryptFunc DecryptFunc) (data []byte, err error)
+}
+
+// DecrypterJSON is ArbitraryDecrypter implementations that can decrypt JSON values
+type DecrypterJSON struct {
+	// Prefix that is required for matching and decrypting values
+	Prefix string
+	// Validate validates data to be decrypted
+	Validate ValidateFunc
```
ondrej-fabry: Then this would be `CheckEncrypted`.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,113 @@
...
+	for k, v := range data {
+		switch t := v.(type) {
+		case string:
+			if s := strings.TrimPrefix(t, d.Prefix); s != t {
+				s, err := base64.URLEncoding.DecodeString(s)
+				if err != nil {
+					return nil, err
+				}
+
+				arbitraryData, err := decryptFunc(s)
```
ondrej-fabry: This should rather be called `decryptedData`.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,113 @@
...
+
+				data[k] = string(arbitraryData)
+			}
+		case map[string]interface{}:
+			v, err := d.decryptJSON(t, decryptFunc)
+			if err != nil {
+				return nil, err
+			}
+
+			data[k] = v
```
ondrej-fabry: Same thing here. You don't need this assignment. Only the on line 100 is needed and will alter the field in the original map.

db/cryptodata/client.go
```diff
@@ -0,0 +1,92 @@
...
+	// Private key is used to decrypt encrypted keys while reading them from store
+	PrivateKeys []*rsa.PrivateKey
+	// Reader used for encrypting/decrypting
+	Reader io.Reader
+	// Hash function used for hashing while encrypting
+	Hash hash.Hash
+}
+
+// ClientWithConfig implements client and client configuration
+type ClientWithConfig struct {
```
ondrej-fabry: This might be too verbose. Let's call this `Client` and the interface `ClientAPI` so it better fits with others.

examples/cryptodata-plugin/main.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return err
+	}
+	encryptedBase64Value := base64.URLEncoding.EncodeToString(encryptedValue)
+	encryptedJSON := fmt.Sprintf(`{"encrypted":true,"value":{"payload":"$crypto$%v"}}`, encryptedBase64Value)
+	plugin.Log.Infof("Putting value %v", encryptedJSON)
+
+	// Put and get
+	key := plugin.etcdKey("value")
+	err = dbWrapped.Put(key, []byte(encryptedJSON))
```
ondrej-fabry: Use just raw `db` for putting the encrypted data to not confuse readers, since `dbWrapped` doesn't actually do anything extra with the data.

Then you can create `dbWrapped` just before the actual `GetValue`.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,111 @@
...
+	"encoding/json"
+	"strings"
+	"encoding/base64"
+)
+
+// DecryptFunc is function that decrypts arbitrary data
+type DecryptFunc func(inData []byte) (data [] byte, err error)
+
+// CheckEncryptedFunc is used to check if JSON data is encrypted
+type CheckEncryptedFunc func(inData []byte) (isValid bool)
```
ondrej-fabry: The return value should be named `isEncrypted`.

examples/cryptodata-plugin/main.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	// Prepare data
+	value := "hello-world"
+	encryptedValue, err := plugin.CryptoData.EncryptData([]byte(value), publicKey)
+	if err != nil {
+		return err
+	}
+	encryptedBase64Value := base64.URLEncoding.EncodeToString(encryptedValue)
+	encryptedJSON := fmt.Sprintf(`{"encrypted":true,"value":{"payload":"$crypto$%v"}}`, encryptedBase64Value)
```
ondrej-fabry: Make this a multi-line string so the contents are more readable.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,111 @@
...
+// CheckEncryptedFunc is used to check if JSON data is encrypted
+type CheckEncryptedFunc func(inData []byte) (isEncrypted bool)
+
+// EncryptionCheck is used to check for data to contain encrypted marker
+type EncryptionCheck struct {
+	// IsEncrypted returns true if data was marked as encrypted
+	IsEncrypted bool `json:"encrypted"`
+}
+
+// ArbitraryDecrypter is interface for decrypting groups of data
```
ondrej-fabry: Should be `ArbitraryDecrypter represents decrypter that looks for encrypted values inside arbitrary data and returns the data with the values decrypted`

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,111 @@
...
+		Prefix: "$crypto$",
+		CheckEncrypted: func(inData []byte) (isValid bool) {
+			var jsonData EncryptionCheck
+			err := json.Unmarshal(inData, &jsonData)
+			return err == nil && jsonData.IsEncrypted
+		},
+	}
+}
+
+// Decrypt tries to decrypt JSON data that are encrypted based on `CheckEncrypted` function return true on data.
```
ondrej-fabry: > Decrypt tries to find encrypted values in JSON data and decrypt them. It uses `CheckEncrypted` function on the data to check if it contains any encrypted data.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,111 @@
...
+type ArbitraryDecrypter interface {
+	// Decrypt decrypts input data using provided decrypting function for arbitrary data
+	Decrypt(inData []byte, decryptFunc DecryptFunc) (data []byte, err error)
+}
+
+// DecrypterJSON is ArbitraryDecrypter implementations that can decrypt JSON values
+type DecrypterJSON struct {
+	// Prefix that is required for matching and decrypting values
+	Prefix string
+	// CheckEncrypted checks if provided data are marked as encrypted
```
ondrej-fabry: Doc for `CheckEncrypted` should describe how the default works.

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,114 @@
...
+
+package cryptodata
+
+import (
+	"encoding/json"
+	"strings"
+	"encoding/base64"
+)
+
+// DecryptFunc is function that decrypts arbitrary data
```
ondrej-fabry: `.. that decrypts input data`

db/cryptodata/decrypter.go
```diff
@@ -0,0 +1,114 @@
...
+// EncryptionCheck is used to check for data to contain encrypted marker
+type EncryptionCheck struct {
+	// IsEncrypted returns true if data was marked as encrypted
+	IsEncrypted bool `json:"encrypted"`
+}
+
+// ArbitraryDecrypter represents decrypter that looks for encrypted values inside arbitrary data and returns
+// the data with the values decrypted
+type ArbitraryDecrypter interface {
+	// Decrypt decrypts input data using provided decrypting function for arbitrary data
```
ondrej-fabry: `Decrypt processes input data and decrypts specific fields using decryptFunc`

PR #314, opened 2018-07-20
Fix trimming prefix for brokers in Consul

ondrej-fabry:
*Note: Consul doesn't allow keys to start with slash: `/`. To support this for our use cases where we use forward slashes in keys, they are stripped from all request before sending to consul.*

This PR fixes cases for `ListValues` and `ListKeys` in which forward slash from key prefix was being trimmed before listing keys/values, but the returned keys from consul were not prepended with the slash, causing iterators fail to strip their original broker prefix (namespace).

### DIFF
--- a/db/keyval/consul/consul.go
+++ b/db/keyval/consul/consul.go
@@ -67,7 +67,7 @@ func NewClient(cfg *api.Config) (store *Client, err error) {
 
 // Put stores given data for the key.
 func (c *Client) Put(key string, data []byte, opts ...datasync.PutOption) error {
-	consulLogger.Debugf("put: %q\n", key)
+	consulLogger.Debugf("Put: %q", key)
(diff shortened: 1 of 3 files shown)


db/keyval/consul/consul.go
```diff
@@ -293,6 +295,7 @@ func (c *Client) watchPrefix(ctx context.Context, prefix string) <-chan watchRes
 					if oldPair, ok := oldPairsMap[pair.Key]; ok {
 						prevVal = oldPair.Value
 					}
+					consulLogger.Warnf(" * modified key: %v prevValue: %v prevModify: %v", pair.Key, len(pair.Value), len(prevVal))
```
VladoLavor: forgotten debug log? Remove it or change to debug

PR #320, opened 2018-07-26
Add support for configuring external hooks for logs

pkotucek:
Signed-off-by: Pavel Kotucek <pkotucek@cisco.com>

### DIFF
--- a/examples/logs-plugin/logs.conf
+++ b/examples/logs-plugin/logs.conf
@@ -7,14 +7,21 @@ loggers:
  - name: "examplechildLogger"
    level: debug
 hooks:
-  fluent:
-    Address: "10.20.30.40"
-    Port: 4521
-    Protocol: tcp
-    Level: debug
-  logstash:
-    Address: "10.20.30.40"
-    Port: 123
-    Protocol: tcp
   syslog:
-    Level: error
+    levels:
+    - panic
+#    - fatal
+#    - error
+#    - warn
+#    - info
+#    - debug
+#  fluent:
+#    address: "10.20.30.41"
+#    port: 4521
+#    protocol: tcp
+#    levels:
+#     - error
+#  logstash:
+#    address: "10.20.30.42"
+#    port: 123
+#    protocol: tcp
--- a/examples/logs-plugin/logs.conf
+++ b/examples/logs-plugin/logs.conf
@@ -10,18 +10,3 @@ hooks:
   syslog:
     levels:
     - panic
-#    - fatal
-#    - error
-#    - warn
-#    - info
-#    - debug
-#  fluent:
-#    address: "10.20.30.41"
-#    port: 4521
-#    protocol: tcp
-#    levels:
-#     - error
-#  logstash:
-#    address: "10.20.30.42"
-#    port: 123
-#    protocol: tcp
--- a/examples/logs-plugin/logs.conf
+++ b/examples/logs-plugin/logs.conf
@@ -10,3 +10,10 @@ hooks:
   syslog:
     levels:
     - panic
+    - error
+    - warn
+#  custom:
+#    addres: localhost
+#    levels:
+#    - panic
+
--- a/examples/logs-plugin/logs.conf
+++ b/examples/logs-plugin/logs.conf
@@ -6,3 +6,16 @@ loggers:
    level: debug
  - name: "examplechildLogger"
    level: debug
+hooks:
+  fluent:
+    Address: localhost
+    Port: 999
+    Protocol: tcp
+  logsys:
+    Address: localhost
+    Port: 4521
+    Protocol: udp
+  logstash:
+    Address: localhost
+    Port: 123
+    Protocol: tcp
--- a/examples/logs-plugin/logs.conf
+++ b/examples/logs-plugin/logs.conf
@@ -8,14 +8,13 @@ loggers:
    level: debug
 hooks:
   fluent:
-    Address: localhost
-    Port: 999
-    Protocol: tcp
-  logsys:
-    Address: localhost
+    Address: "10.20.30.40"
     Port: 4521
-    Protocol: udp
+    Protocol: tcp
+    Level: debug
   logstash:
-    Address: localhost
+    Address: "10.20.30.40"
     Port: 123
     Protocol: tcp
+  syslog:
+    Level: error
--- a/logging/logmanager/plugin_impl_log_manager.go
+++ b/logging/logmanager/plugin_impl_log_manager.go
@@ -128,7 +128,7 @@ func (lm *Plugin) Init() error {
 					logCfgEntry.Name, err)
 			}
 		}
-		lm.Log.Warn("configuring log hooks")
+		lm.Log.Info("configuring log hooks")
 		for hookName, hookConfig := range lm.Conf.Hooks {
 			if err := lm.addHook(hookName, hookConfig); err != nil {
 				lm.Log.Warnf("configuring log hook %s failed: %v", hookName, err)
@@ -309,7 +309,7 @@ func (lm *Plugin) addHook(hookName string, hookConfig HookConfig) error {
 		}
 	}
 	// add hook to existing loggers and store it into registry for late use
-	lm.Log.Warnf("add hook %v to registry", hookName)
+	lm.Log.Infof("add hook %v to registry", hookName)
 	lm.LogRegistry.AddHook(cHook)
 	return nil
 }
--- a/logging/logmanager/plugin_impl_log_manager.go
+++ b/logging/logmanager/plugin_impl_log_manager.go
@@ -21,14 +21,14 @@ import (
 	"os"
 	"strings"
 
-	lLogStash "github.com/bshuster-repo/logrus-logstash-hook"
-	lFluent "github.com/evalphobia/logrus_fluent"
+	"github.com/bshuster-repo/logrus-logstash-hook"
+	"github.com/evalphobia/logrus_fluent"
 	"github.com/gorilla/mux"
 	"github.com/ligato/cn-infra/infra"
 	"github.com/ligato/cn-infra/logging"
 	"github.com/ligato/cn-infra/rpc/rest"
 	"github.com/sirupsen/logrus"
-	lSyslog "github.com/sirupsen/logrus/hooks/syslog"
+	lgSyslog "github.com/sirupsen/logrus/hooks/syslog"
 	"github.com/unrolled/render"
 	"strconv"
 )
@@ -96,7 +96,6 @@ func (lm *Plugin) Init() error {
 		if err != nil {
 			return err
 		}
-		lm.Log.Debugf("logs config: %+v", lm.Conf)
 
 		// Handle default log level. Prefer value from environmental variable
 		defaultLogLvl := os.Getenv("INITIAL_LOGLVL")
@@ -129,8 +128,12 @@ func (lm *Plugin) Init() error {
 					logCfgEntry.Name, err)
 			}
 		}
-		lm.Log.Warn("hooks config ... ")
+		lm.Log.Warn("configuring log hooks ...")
 
+		createLevelMappings()
+		lm.AddHook(HookSysLog)
+		lm.AddHook(HookLogStash)
+		lm.AddHook(HookFluent)
 	}
 
 	return nil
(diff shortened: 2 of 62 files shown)


logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -16,15 +16,21 @@ package logmanager
 
 import (
 	"fmt"
+	"log/syslog"
 	"net/http"
 	"os"
 	"strings"
 
+	lLogStash "github.com/bshuster-repo/logrus-logstash-hook"
```
ondrej-fabry: Better to have `snake_case` than `camelCase` here.

VladoLavor: I'm just curious why, because personally I prefer camel case. Also IDE advises the opposite, referencing this https://golang.org/doc/effective_go.html#mixed-caps

ondrej-fabry: Actually I would go for no alias at all.

And [Effective Go] actually advises this for packages:
> By convention, packages are given **lower case**, **single-word names**; there should be no **need for underscores or mixedCaps**
https://golang.org/doc/effective_go.html#package-names

ondrej-fabry: @pkotucek Why did you feel the need to use aliases at all? They should be used only if there are some collisions.

pkotucek: I removed them.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +229,88 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+		logging.PanicLevel: {logrus.PanicLevel},
+		logging.FatalLevel: {logrus.FatalLevel},
+		logging.ErrorLevel: {logrus.ErrorLevel},
+		logging.WarnLevel:  {logrus.WarnLevel},
+		logging.InfoLevel:  {logrus.InfoLevel},
+		logging.DebugLevel: {logrus.DebugLevel},
+	}
+}
+
+func (lm *Plugin) AddHook(hookName string) {
```
ondrej-fabry: Seems like this should return error.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +229,88 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+	Address  string
+	Port     int
+	Level    string
+}
+
+var llm = levelMapping{}
+
+func createLevelMappings() {
+	llm.syslogLevel = map[logging.LogLevel]syslog.Priority{
+		logging.PanicLevel: syslog.LOG_CRIT,
```
ondrej-fabry: This kind of conversion already happens inside logrus's syslog hook package: https://github.com/sirupsen/logrus/blob/master/hooks/syslog/syslog.go#L35-L49

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +229,88 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+		logging.InfoLevel:  {logrus.InfoLevel},
+		logging.DebugLevel: {logrus.DebugLevel},
+	}
+}
+
+func (lm *Plugin) AddHook(hookName string) {
+	var hook logrus.Hook
+	var err error
+
+	if hookConfig, exists := lm.Conf.Hooks[hookName]; exists {
```
ondrej-fabry: This should be processed in Init.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -87,7 +96,6 @@ func (lm *Plugin) Init() error {
 		if err != nil {
 			return err
 		}
-		lm.Log.Debugf("logs config: %+v", lm.Conf)
```
ondrej-fabry: Removing this why?

pkotucek: Ups, deleted by mistake with my temp debugs.

logging/logmanager/plugin_impl_log_manager.go
pkotucek: I need these conversions when creating hooks.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +229,88 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+type levelMapping struct {
+	syslogLevel map[logging.LogLevel]syslog.Priority
+	fluentLevel map[logging.LogLevel][]logrus.Level
+}
+
+type HookConfig struct {
+	Protocol string
+	Address  string
+	Port     int
+	Level    string
```
ondrej-fabry: We will want slice here. There should be option to set for example both error and warning for hooks.

logging/logmanager/plugin_impl_log_manager.go
ondrej-fabry: Have you tested how it currently behaves? I believe that if I set hook's level to `panic` now, it will log with `syslog.LOG_CRIT` to syslog.. but it will log ALL levels! 
See my latest review.

ondrej-fabry: Please, if you have trouble grasping how logrus behaves. Create small separate project to test out all combinations until you are certain with its behavior. Only then continue with implementation here.

examples/logs-plugin/logs.conf
```diff
@@ -6,3 +6,7 @@ loggers:
    level: debug
  - name: "examplechildLogger"
    level: debug
+hooks:
+  syslog:
+    levels:
+    - panic
```
ondrej-fabry: Example should show use case with multiple levels.

ondrej-fabry: Ideally, panic/fatal/error, since first two of those will stop the program.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +234,91 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+		return fmt.Errorf("unsupported hook")
+	}
+	// create hook
+	cHook = &CommonHook{
+		lgHook,
+		[]logrus.Level{},
+	}
+	// fill up defined levels, or use default if not defined
+	if len(hookConfig.Levels) == 0 {
+		lgl, _ := logrus.ParseLevel(lm.Conf.DefaultLevel)
```
ondrej-fabry: This is not the good default. It means that if not levels are defined, only info level will be sent to the hook. The useful default for hooks is having levels: panic/fatal/error, and maybe warn but probably not.

examples/logs-plugin/logs.conf
pkotucek: An example is in cn-infra/logging/logging.conf

logging/logmanager/plugin_impl_log_manager.go
pkotucek: So we can introduce default level for hook in config, can we?

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +234,91 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+		cHook.levels = append(cHook.levels, lgl)
+	} else {
+		for _, level := range hookConfig.Levels {
+			lgl, _ := logrus.ParseLevel(level)
```
ondrej-fabry: Why ignore the error? We want this to fail if invalid level is defined.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +234,91 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+			cHook.levels = append(cHook.levels, lgl)
+		}
+	}
+	// add hook to existing loggers and store it into registry for late use
+	if err == nil {
```
ondrej-fabry: The error should be checked right after the switch above.

logging/logrus/registry.go
```diff
@@ -229,3 +231,8 @@ func (lr *logRegistry) getLoggerFromMapping(logger string) *Logger {
 	panic("cannot cast log value to Logger obj")
 
 }
+
+// HookConfigs stores hook configs from log manager
+func (lr *logRegistry) AddHook(hook logrus.Hook) {
+	lr.hooks = append(lr.hooks, hook)
```
ondrej-fabry: Where is the hooks field even used? It should be applied to all new loggers created in this registry.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +234,91 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+	case HookFluent:
+		lgHook, err = logrus_fluent.NewWithConfig(logrus_fluent.Config{
+			Host: hookConfig.Address,
+			Port: hookConfig.Port,
+		})
+	default:
+		return fmt.Errorf("unsupported hook")
+	}
+	// create hook
+	cHook = &CommonHook{
```
ondrej-fabry: You don't have to define this variable at the top of the function, you can just use `cHook := &CommonHook{`.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +234,91 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+// HookConfig contains configuration of hook services
+type HookConfig struct {
+	Protocol string
+	Address  string
+	Port     int
+	Levels   []string
+}
+
+// CommonHook implements that Hook with own level definition
+type CommonHook struct {
```
ondrej-fabry: No need to export this type, it's not gonna be useful for anyone outside this package.

examples/logs-plugin/logs.conf
ondrej-fabry: That's not example of program.. it's an example for configuration. I want to run this example to see how the hooks work and what I can do with them.

Also, perhaps new example should be defined for hooks. Please copy this one and name it `logs-hooks-plugin`.

logging/logmanager/plugin_impl_log_manager.go
ondrej-fabry: I would rather not complicate it. Let's just put panic/fatal/error as default if no levels are defined for hook.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -120,7 +128,12 @@ func (lm *Plugin) Init() error {
 					logCfgEntry.Name, err)
 			}
 		}
-
+		lm.Log.Info("configuring log hooks")
```
ondrej-fabry: This can be just debug level or you could put all this into if len(lm.Conf.Hooks) > 0

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -120,7 +128,12 @@ func (lm *Plugin) Init() error {
...
+		for hookName, hookConfig := range lm.Conf.Hooks {
+			if err := lm.addHook(hookName, hookConfig); err != nil {
+				lm.Log.Warnf("configuring log hook %s failed: %v", hookName, err)
```
ondrej-fabry: This should be error.. and should `return err`.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +229,87 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+	}
+	// create hook
+	cHook := &commonHook{
+		lgHook,
+		[]logrus.Level{},
+	}
+	// fill up defined levels, or use default if not defined
+	var levels []string
+	if len(hookConfig.Levels) == 0 {
+		levels = []string{ "fatal", "panic", "error" }
```
ondrej-fabry: You can set `cHook.levels = []logrus.Level{logrus.Fatal, logrus.Panic...` instead of using strings and then parsing them.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +229,87 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+	}
+	for _, level := range levels {
+		if lgl, err := logrus.ParseLevel(level); err == nil {
+			cHook.levels = append(cHook.levels, lgl)
+		} else {
+			lm.Log.Warnf("couldn't parse level %v : %v", level, err.Error())
+		}
+	}
+	// add hook to existing loggers and store it into registry for late use
+	lm.Log.Infof("add hook %v to registry", hookName)
```
ondrej-fabry: This should be inside the registry's AddHook method, so it's showed for those who might set it directly.

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +229,87 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+	case HookFluent:
+		lgHook, err = logrus_fluent.NewWithConfig(logrus_fluent.Config{
+			Host: hookConfig.Address,
+			Port: hookConfig.Port,
+		})
+	default:
+		return fmt.Errorf("unsupported hook")
+	}
+	if err != nil {
+		lm.Log.Warnf("couldn't create hook for %v : %v", hookName, err.Error())
```
ondrej-fabry: This will print two warnings when error occurs. You can rather return custom error: `return fmt.Errorf("creating hook for %v failed: %v", hookName, err)`

logging/logmanager/plugin_impl_log_manager.go
```diff
@@ -216,3 +229,87 @@ func stringToLogLevel(level string) logging.LogLevel {
...
+		return err
+	}
+	// create hook
+	cHook := &commonHook{
+		lgHook,
+		[]logrus.Level{},
```
ondrej-fabry: No need to set this to anything. You can just use `&commonHook{Hook: lgHook}`

logging/logrus/registry.go
```diff
@@ -229,3 +237,17 @@ func (lr *logRegistry) getLoggerFromMapping(logger string) *Logger {
...
+// HookConfigs stores hook configs provided by log manager
+// and applies hook to existing loggers
+func (lr *logRegistry) AddHook(hook logrus.Hook) {
+	lr.hooks = append(lr.hooks, hook)
+
+	lgs := lr.ListLoggers()
+	for lg := range lgs {
+		logger, found := lr.Lookup(lg)
+		if found {
+			logger.AddHook(hook)
```
ondrej-fabry: Are you sure this works?? Because I don't see implementation for this...

logging/logmanager/plugin_impl_log_manager.go
ondrej-fabry: It's still removed..

logging/logrus/registry.go
pkotucek: There is implementation in logger.go
func (logger *Logger) AddHook(hook lg.Hook)

logging/logmanager/plugin_impl_log_manager.go
ondrej-fabry: You most likely didn't define it as I wrote.

PR #324, opened 2018-08-02
example cleanup

VladoLavor:

### DIFF
--- a/examples/datasync-plugin/main.go
+++ b/examples/datasync-plugin/main.go
@@ -28,65 +28,28 @@ import (
 // ************************************************************************/
 
 // PluginName represents name of plugin.
-const PluginName = "example"
+const PluginName = "datasync-example"
 
 func main() {
-	// Start Agent with ExamplePlugin, ETCDPlugin & FlavorLocal (reused cn-infra plugins).
-	/*agent := local.NewAgent(local.WithPlugins(func(flavor *local.FlavorLocal) []*core.NamedPlugin {
-		etcdPlug := &etcd.Plugin{}
-		etcdDataSync := &kvdbsync.Plugin{}
-		resyncOrch := &resync.Plugin{}
-
-		etcdPlug.Deps.PluginInfraDeps = *flavor.InfraDeps("etcd", local.WithConf())
-		resyncOrch.Deps.PluginLogDeps = *flavor.LogDeps("etcd-resync")
-		connectors.InjectKVDBSync(etcdDataSync, etcdPlug, etcdPlug.PluginName, flavor, resyncOrch)
-
-		examplePlug := &ExamplePlugin{closeChannel: exampleFinished}
-		examplePlug.Deps.PluginInfraDeps = *flavor.InfraDeps("etcd-example")
-		examplePlug.Deps.Publisher = etcdDataSync // Inject datasync Watcher to example plugin.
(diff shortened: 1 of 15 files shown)


examples/datasync-plugin/main.go
```diff
@@ -28,65 +28,28 @@ import (
...
-		return []*core.NamedPlugin{
-			{etcdPlug.PluginName, etcdPlug},
-			{etcdDataSync.PluginName, etcdDataSync},
-			{resyncOrch.PluginName, resyncOrch},
-			{examplePlug.PluginName, examplePlug},
-		}
+	// Prepare ETCD data sync plugin as an plugin dependency
+	etcdDs := kvdbsync.NewPlugin(kvdbsync.UseDeps(func(deps *kvdbsync.Deps) {
+		deps.KvPlugin = &etcd.DefaultPlugin
+		deps.ResyncOrch = &resync.DefaultPlugin
```
ondrej-fabry: The resync was recently added as default dependency, since datasync is only ever used with it. You don't have to specifically set it. For the KV store, there is special option `UseKV` that makes all of this much shorter and cleaner.

examples/datasync-plugin/main.go
```diff
@@ -28,65 +28,28 @@ import (
...
-
-	p := &ExamplePlugin{
+	// Init example plugin dependencies
+	ep := &ExamplePlugin{
 		Deps: Deps{
 			Log:          logging.ForPlugin(PluginName),
 			ServiceLabel: &servicelabel.DefaultPlugin,
-			Publisher:    etcdDataSync,
-			Watcher:      etcdDataSync,
+			Publisher:    etcdDs,
```
ondrej-fabry: We should probably keep the longer version to make this more clear to readers of example.

examples/kafka-plugin/hash-partitioner/main.go
```diff
@@ -186,7 +193,7 @@ func (plugin *ExamplePlugin) closeExample() {
 
 			plugin.Log.Info("kafka example finished, sending shutdown ...")
 
-			plugin.closeChannel <- struct{}{}
+			plugin.exampleFinished <- struct{}{}
```
ondrej-fabry: Did this example even stop automatically? Because the `exampleFinished` should be closed and not send data to it.

ondrej-fabry: Never mind.. realized the channel doesn't differ between send and close... which to me seems more like unintended behavior.

examples/redis-plugin/main.go
```diff
@@ -1,57 +1,70 @@
...
-
-		return []*core.NamedPlugin{
-			{redisPlug.PluginName, redisPlug},
-			{redisDataSync.PluginName, redisDataSync},
-			{resyncOrch.PluginName, resyncOrch},
-			{examplePlug.PluginName, examplePlug}}
+	// Prepare Redis data sync plugin as an plugin dependency
+	redisDs := kvdbsync.NewPlugin(kvdbsync.UseDeps(func(deps *kvdbsync.Deps) {
+		deps.KvPlugin = &etcd.DefaultPlugin
+		deps.ResyncOrch = &resync.DefaultPlugin
```
ondrej-fabry: Same thing with default and `UseKV` option.

PR #326, opened 2018-08-06
KVscheduler

milanlenco:
What KVScheduler is still missing is:
 - docs, developer guide in particular


kvscheduler/descriptor/adapter-generator/template.go
```diff
@@ -0,0 +1,353 @@
+package main
+
+const Template = `
+// Code generated by adapter-generator. DO NOT EDIT.
+// This file was generated by at {{ .Timestamp }}
```
ondrej-fabry: Generated file should not contain timestamp, otherwise it will be harder to do checks if the generated file changed. The file itself has modification timestamp.

kvscheduler/plugin_scheduler.go
```diff
@@ -0,0 +1,296 @@
...
+
+const (
+	// DependencyRelation identifies dependency relation for the graph.
+	DependencyRelation = "depends-on"
+
+	// DerivesRelation identifies relation of value derivation for the graph.
+	DerivesRelation = "derives"
+)
+
+var schedulerSingleton *Scheduler
```
ondrej-fabry: Why use singleton? Seems like this should be generic plugin like others and be as dependency for others.

idxmap/api.go
```diff
@@ -34,14 +34,18 @@ type NamedMapping interface {
...
 	//
 	// Example usage:
 	//
-	//  map.Watch(plugin.PluginName, ToChan(myChannel))
+	//  map.Watch(subscriber, ToChan(myChannel))
 	//
-	//  map.Watch(plugin.PluginName, func(msgNamedMappingGenericEvent) {/*handle callback*/ return nil})
+	//  map.Watch(subscriber, func(msgNamedMappingGenericEvent) {/*handle callback*/ return nil})
 	//
 	Watch(subscriber infra.PluginName, callback func(NamedMappingGenericEvent)) error
```
ondrej-fabry: What blocks us from removing dependency on `infra` package? It should be enough for subscriber to be string.

kvscheduler/descriptor/adapter-generator/template.go
```diff
@@ -0,0 +1,365 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+const Template = `
```
ondrej-fabry: Extra newline at the first line can be removed here.

kvscheduler/descriptor/adapter-generator/template.go
```diff
@@ -0,0 +1,365 @@
...
+{{- end }}
+}
+{{- else }}
+func (db *{{ .DescriptorName }}DescriptorBase) Build(key string, valueData {{ .ValueDataT }}) (value {{ .ValueT }}, err error) {
+	return nil, nil
+}
+{{- end }}
+
+func (db *{{ .DescriptorName }}DescriptorBase) Add(key string, value {{ .ValueT }}) (metadata {{ .MetadataT }}, err error) {
+	fmt.Printf("Create for key=%s is not implemented\n", key)
```
ondrej-fabry: Instead of printing `not implemented` message, we could return an error that it is not implemented, otherwise the callers of the method will assume it succeeded.

kvscheduler/descriptor/adapter-generator/template.go
```diff
@@ -0,0 +1,365 @@
...
+}
+
+func (db *{{ .DescriptorName }}DescriptorBase) DumpDependencies() []string {
+	return nil
+}
+
+////////// Descriptor adapter //////////
+
+type {{ .DescriptorName }}DescriptorAdapter struct {
+	descriptor {{ .DescriptorName }}DescriptorAPI
```
ondrej-fabry: By using anonymous field you could omit most of the boiler-plate below and only keep those methods which actually provide some override.

kvscheduler/descriptor/adapter-generator/template.go
milanlenco: Actually it is OK not to implement a descriptor method. The printf statements I used only for testing, so I guess they can be removed.

kvscheduler/descriptor/adapter-generator/template.go
milanlenco: good point, will do

idxmap/api.go
milanlenco: I actually planned to changed that to string but got distracted and forgot about it - will do with the rest of the review comments.

kvscheduler/descriptor/adapter-generator/template.go
ondrej-fabry: You mean like this? https://play.golang.org/p/AGDP_nfhisL

ondrej-fabry: My terminology is definitely not very correct, because it's not actually override like in C++ classes. They just

https://golang.org/ref/spec#Method_sets
https://golang.org/ref/spec#Selectors

milanlenco: cool, will use the anonymous field then

kvscheduler/api/kv_scheduler_api.go
```diff
@@ -0,0 +1,319 @@
...
+)
+
+// String converts ValueType to string.
+func (vt ValueType) String() string {
+	switch vt {
+	case Object:
+		return "object"
+	case Action:
+		return "action"
+	default:
```
ondrej-fabry: This case should match if `vt==Property` and use default for all other values. Otherwise something like this could occur:

```go
v := ValueType(-1)
v.String() // will print "property"
if v == Property {
 // will not be executed
}
```

This might be harder to debug.

milanlenco:
Merging to devel branch. Review comments will be reflected when linux (and maybe also vpp-ifplugin) are refactored.

Summary of changes TBD from review :
- remove pluginName from idxmap
- remove newline from the start of the template 
- move descriptor from interface to structure - nil parameter = use default (consider if not implemented action is error or not)
- add isRetriable(error) to the descriptor
- rename NBKeyPrefixes -> to GetNBKeyPrefixes
- get rid of Value, use just proto.Message - descriptor still needs to be able to override the default Equal method()
[DONE]- get rid of ValueType - no use for it (derived values will be allowed to be without descriptor)
- rename Copy to DeepCopy in KeySet
- remove logs from REST registration
- transaction handling: check we why continue when values == 0 ?
- resync transaction entered usind SetValueData (get rid of Resync())
- rename ModifyHasToReplace to ModifyWithReplace
- cancel transaction on any validation error

Issue #327, opened 2018-08-13
Use Dependabot for cn-infra


mestery:
We recently moved to using Dependabot [1] for Network Service Mesh, and it's pretty slick. I think using it for cn-infra (and vpp-agent, sfc-controller, etc.) makes a lot of sense. It's easy to add, keeps your dependencies up to date, and is a smooth workflow.

[1] https://dependabot.com/

ondrej-fabry:
> Dependabot doesn’t yet support vendoring for Go dependencies so if you commit a vendor folder Dependabot won’t update it for you.

https://dependabot.com/blog/go-support

Not feasible for us at the moment, we commit vendor folder with dependencies.

mestery:
Thanks for pointing this out, and it makes sense after seeing the @dependabot pushes to the Network Service Mesh repository. However, one thing you can do in the meantime is let Dependabot handle pushing the Gopkg.[lock,toml] updates, and then pulling that down and running `dep ensure -update` and pushing back to the @Dependabot PR. It's a workaround until they implement full vendoring support, and is how we're doing this for NSM.

ondrej-fabry:
This sounds plausible if it were run in some script via Travis so it would do this work for open PRs by Dependabot.

Are you sure you are actively doing that for networkservicemesh?

Running `dep check` for the latest networkservicemehs returns this:

```sh
# Gopkg.lock is out of sync:
github.com/golang/protobuf/proto: imported or required, but missing from Gopkg.lock's input-imports
github.com/golang/protobuf/protoc-gen-go: imported or required, but missing from Gopkg.lock's input-imports
github.com/ligato/cn-infra/config: imported or required, but missing from Gopkg.lock's input-imports
github.com/ligato/cn-infra/core: imported or required, but missing from Gopkg.lock's input-imports
github.com/ligato/cn-infra/flavors/local: imported or required, but missing from Gopkg.lock's input-imports
github.com/ligato/cn-infra/flavors/rpc: imported or required, but missing from Gopkg.lock's input-imports
github.com/ligato/cn-infra/health/statuscheck: imported or required, but missing from Gopkg.lock's input-imports
github.com/ligato/cn-infra/logging: imported or required, but missing from Gopkg.lock's input-imports
golang.org/x/net/context: imported or required, but missing from Gopkg.lock's input-imports
google.golang.org/grpc: imported or required, but missing from Gopkg.lock's input-imports
[...]

PR #328, opened 2018-08-13
Update dependencies

mestery:
This commit updates all of the dependencies for cn-infra and get the
entire package compiling (as shown by running `make` from the
toplevel).

Note this does upgrade etcd to version 3.3.9, because version 3.2.0
no longer built with updated dependencies.

Please note this is an attempt to bring the dependencies up to date outside of the work being down around adding in requisite dependencies for the custom resource definition work being done.


mestery:
@ondrej-fabry @jmedved @jhnbrns FYI, this seems to update all cn-infra dependencies and get things compiling. Is there a reason these are not kept up to date? See #327 for an easy way to keep this up to date automatically.

mestery:
I am unsure why `make check-links` fails, seems unrelated to my change. The `make test-cover` failure is due to the etcd update, though it's not clear to me why that fails either.

mestery:
This is passing CI now, I'd like to have this on merged and then I can move on to updating the dependencies in vpp-agent and finally contiv-vpp.

lukasmacko:
I would recommand to try it with vpp-agent and contiv/vpp first. It can be tested using [source rule](https://golang.github.io/dep/docs/Gopkg.toml.html#source) in `gopkg.toml` that will point to your fork with updated dependencies.

Once we know that the new versions work for the projects this could be merged.

ondrej-fabry:
As @lukasmacko already wrote. You should try that these version are compatible with projects that use cn-infra, mainly vpp-agent and contiv. The specific versions in Gopkg.toml were [chosen](https://github.com/ligato/cn-infra/commit/5d7f47939ce484d71d0b7b58f725f676a6e4a43a#diff-836546cc53507f6b2d581088903b1785) by @lukasmacko due some dependency issues when using cn-infra in contiv.

mestery:
Thanks @lukasmacko and @ondrej-fabry, I was testing with the source rule, will keep at it. BTW, I could only test this with vpp-agent with the contiv-1807 branch there, is there plans to merge that branch into master anytime soon?

ondrej-fabry:
@mestery For vpp-agent, we do main development on pantheon-dev branch. Tomorrow planning to merge that into contiv-1807, which is special branch just for contiv/vpp project. Master is used for official releases into which we merge from pantheon-dev.

Gopkg.toml
```diff
@@ -25,16 +25,17 @@
 #   unused-packages = true
 
 required = [
-    "github.com/grpc-ecosystem/grpc-gateway/runtime"
+    "github.com/grpc-ecosystem/grpc-gateway/runtime",
+    "github.com/coreos/bbolt"
 ]
 
-[[constraint]]
+[[override]]
```
ondrej-fabry: Why is this override? It's direct dependency for cn-infra: https://github.com/ligato/cn-infra/blob/master/rpc/grpc/plugin_impl_grpc.go#L25

Gopkg.toml
```diff
@@ -54,7 +55,11 @@ required = [
 
 [[constraint]]
   name = "github.com/coreos/etcd"
-  version = "=3.2.0"
+  version = "=3.3.9"
+
+[[constraint]]
+  name = "github.com/coreos/bbolt"
```
ondrej-fabry: When did we switch to bbolt and why?

mestery: Something else was pulling bbolt and I had to add this constraint, otherwise a `dep ensure -update` resulted in a tree which failed to compile.

Gopkg.toml
mestery: I recall @edwarnicke indicating there as a problem with cn-infra and newer grpc versions. The override here allows this to bump grpc up to version 1.14.0 (the latest). If this will be a problem, let me know.

Gopkg.toml
ondrej-fabry: Could you share the failure error? We use [Bolt](https://github.com/boltdb/bolt) and not [BBolt](https://github.com/coreos/bbolt) so I'm really curious what was *pulling* this dependency. There is [no mention of bbolt](https://github.com/ligato/cn-infra/search?q=bbolt&unscoped_q=bbolt) anywhere in cn-infra.

mestery: I just pulled this out and re-ran `dep ensure -update`, and dep completed. So now I'm confused. I'l push a commit to remove it explicitly from Gopkg.toml. However, note that bbolt continues to remain in vendor, I believe because etcd is using it:

```
machine:cn-infra userid$ grep -R bbolt vendor/
vendor//github.com/boltdb/bolt/README.md:> you look at the CoreOS fork called [bbolt](https://github.com/coreos/bbolt).
grep: vendor//github.com/hashicorp/go-rootcerts/test-fixtures/capath-with-symlinks/securetrust.pem: No such file or directory
grep: vendor//github.com/hashicorp/go-rootcerts/test-fixtures/capath-with-symlinks/thawte.pem: No such file or directory
grep: vendor//github.com/coreos/etcd/cmd/tools: No such file or directory
grep: vendor//github.com/coreos/etcd/cmd/etcdctl: No such file or directory
grep: warning: vendor//github.com/coreos/etcd/cmd/etcd: recursive directory loop
grep: vendor//github.com/coreos/etcd/cmd/functional: No such file or directory
grep: vendor//github.com/coreos/etcd/Documentation/README.md: No such file or directory
vendor//github.com/coreos/etcd/mvcc/backend/backend.go:	bolt "github.com/coreos/bbolt"
vendor//github.com/coreos/etcd/mvcc/backend/batch_tx.go:	bolt "github.com/coreos/bbolt"
vendor//github.com/coreos/etcd/mvcc/backend/config_windows.go:import bolt "github.com/coreos/bbolt"
vendor//github.com/coreos/etcd/mvcc/backend/read_tx.go:	bolt "github.com/coreos/bbolt"
vendor//github.com/coreos/etcd/mvcc/backend/config_default.go:import bolt "github.com/coreos/bbolt"
vendor//github.com/coreos/etcd/mvcc/backend/config_linux.go:	bolt "github.com/coreos/bbolt"
machine:cn-infra userid$ 
```

ondrej-fabry: The error probably wasn't occurring during dep, but during running tests. Just like it's happening in Travis now. 
I believe that the reason is that etcd project uses glide and might have issue figuring out bbolt dependency.

You selected etcd v3.3.9 which [uses some specific option](https://github.com/coreos/etcd/blob/v3.3.9/mvcc/backend/config_linux.go#L31) from bbolt [introduced in bbolt v1.3.1](https://github.com/coreos/bbolt/blob/v1.3.1-coreos.6/db.go#L1003), although the dep chose v1.3.0 causing tests to fail to compile.

This means you should actually keep the bbolt in `Gopkg.toml`. However use `[[override]]` since we don't actually use bbolt directly, only as transitive dependency from etcd.

Gopkg.toml
```diff
@@ -28,13 +28,13 @@ required = [
     "github.com/grpc-ecosystem/grpc-gateway/runtime"
 ]
 
-[[constraint]]
+[[override]]
```
ondrej-fabry: This is still relevant to my comment: https://github.com/ligato/cn-infra/pull/328#discussion_r211009794

Gopkg.toml
```diff
@@ -130,7 +130,7 @@ required = [
 
 [[override]]
   name = "github.com/ugorji/go"
-  revision = "ded73eae5db7e7a0ef6f55aace87a2873c5d2b74"
+  revision = "v1.1.1"
```
ondrej-fabry: This should most likely be `version = "1.1.1"`.

Gopkg.lock
```diff
@@ -2,60 +2,85 @@
...
-  version = "v2.1.11"
+  pruneopts = "UT"
+  revision = "c618e605e15c0d7535f6c96ff8efbb0dba4fd66c"
+  version = "v2.1.15"
 
 [[projects]]
-  branch = "master"
-  name = "github.com/cockroachdb/cmux"
+  digest = "1:c28625428387b63dd7154eb857f51e700465cfbf7c06f619e71f2da33cefe47e"
+  name = "github.com/coreos/bbolt"
```
ondrej-fabry: See: https://github.com/ligato/cn-infra/pull/328#discussion_r211144328

Gopkg.toml
mestery: 👍

Gopkg.toml
mestery: 👍

Gopkg.toml
ondrej-fabry: You can't really compare output of a `dep status` with syntax used in `Gopkg.toml` file.

My initial comment was more about coming from official documentation for `Gopkg.toml` and it's `version` property: https://golang.github.io/dep/docs/Gopkg.toml.html#version

And if it allows having both cases. Why go for inconsistency if there is no other `version` property with `v` prefix in current `Gopkg.toml`. Inconsistency creates confusion. Either put this without prefix or put all others with prefix.

mestery: That's a good explanation and I'll fix this one. Thanks @ondrej-fabry!

mestery: OK, something weird happened. When I changes this to `1.1.1` instead of `v1.1.1`, a run of `dep ensure -update` failed because it could no longer match the version of `github.com/ugorji/go` it needed for etcd-3.3.9. I wonder if this is related to their usage of glide.

ondrej-fabry:
Rebase after https://github.com/ligato/cn-infra/pull/331 

Also, remove `github.com/coreos/bbolt` from required and `github.com/ugorji/go` from `Gopkg.toml` that helped me when trying locally and running `dep ensure`.

PR #340, opened 2018-09-18
stopwatch replaced with trace

VladoLavor:


logging/measure/tracer.go
```diff
@@ -0,0 +1,133 @@
...
+func (t *tracer) Start() {
+	t.startTime = time.Now()
+}
+
+func (t *tracer) LogTime(entity string) {
+	if t == nil {
+		return
+	}
+	// Skip cli-inband
+	if entity == "cli_inband" {
```
ondrej-fabry: Why is this here?

logging/measure/tracer.go
```diff
@@ -0,0 +1,133 @@
...
+	t.timedb = make(map[int]*entry)
+}
+
+func (t *tracer) process() *apitrace.Trace {
+	t.Lock()
+	defer t.Unlock()
+
+	var (
+		trace = &apitrace.Trace{
+			TracedEntries: make([]*apitrace.Trace_TracedEntry, 0),
```
ondrej-fabry: This will get overridden at the bottom anyway.

logging/measure/tracer.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// Single time entry
+type entry struct {
+	entryName  string
+	loggedTime time.Duration
+}
+
+func (t *tracer) Start() {
+	t.startTime = time.Now()
```
ondrej-fabry: All entries should not share same start time. What if two entries are being logged at the same time?

logging/measure/tracer.go
VladoLavor: added it bc is spammed a lot, and forget to remove, thanks

VladoLavor:
returned average values (I was not sure about the usefulness), anything else is missing? Runtime logging was intentionally removed since it just polluted the log and was hard to find a specific value.

logging/measure/tracer.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Inner structure handling database and measure results
+type tracer struct {
+	sync.Mutex
+
+	name string
+	log  logging.Logger
+	// Entry index, used in database as key and increased after every entry. Never resets since the tracer object is
+	// created or the database is cleared
+	index int
```
ondrej-fabry: This should be `uint64`.

logging/measure/model/apitrace/apitrace.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+// Package apitrace provides data model for binary API trace informations.
+package apitrace;
+
+message Trace {
+    message TracedEntry {
+        uint32 index = 1;                       // Trace index
+        string msg_name = 2;                    // Message name (or any other string)
+        string duration = 3;                    // Duration in milliseconds
```
ondrej-fabry: Not sure if it's good idea to keep duration as string. It's gonna make it harder for parse the value for external systems. It should be integer, perhaps with unit as part of name so it's clear to readers.

VladoLavor: true, but I would keep both formats, the integer which is easy to parse, and keep also the sting which is easier to read

ondrej-fabry: I'd strongly oppose this. It might easily cause problems when these two values are not in sync and also very likely to confuse readers. 

I think consumers of this data are gonna do what they need with the value anyway, but generally we should provide low level (unsigned) integer value which is also easily comparable.

ondrej-fabry: Perhaps this might be useful? https://godoc.org/github.com/gogo/protobuf/types#Duration

logging/measure/tracer.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Inner structure handling database and measure results
+type tracer struct {
+	sync.Mutex
+
+	name string
+	log  logging.Logger
+	// Entry index, used in database as key and increased after every entry. Never resets since the tracer object is
+	// created or the database is cleared
+	index uint32
```
ondrej-fabry: This better be `uint64` as I suggested [before](https://github.com/ligato/cn-infra/pull/340/#discussion_r218747362). We don't currently have any limit and clear doesn't reset this index so we want to go for largest possible integer. Change the type of index in the model too.

logging/measure/tracer.go
```diff
@@ -0,0 +1,135 @@
...
+	})
+	t.index++
+}
+
+func (t *tracer) Get() *apitrace.Trace {
+	t.Lock()
+	defer t.Unlock()
+
+	var (
+		average map[string][]time.Duration // message name -> measured times
```
ondrej-fabry: You could just create instance of `*apitrace.Trace` here instead of creating additional variables and then setting them for fields at the end, since you are not using `return nil` anyway there won't be any unnecessary allocation and actually not that many variables.

logging/measure/tracer.go
```diff
@@ -0,0 +1,135 @@
...
+	average = make(map[string][]time.Duration)
+
+	for _, entry := range t.timedb {
+		// Add to total
+		total += entry.loggedTime
+		// Add to message data
+		message := &apitrace.Trace_Entry{
+			Index:    entry.index,
+			MsgName:  entry.name,
+			Duration: uint32(entry.loggedTime.Nanoseconds()),
```
ondrej-fabry: Be careful here! The maximum value of `uint32` is `4294967295` and that is only ~4 seconds in nanoseconds, which time unit of `time.Duration`. This means any duration longer than that the `uint32` will silently overflow and start showing wrong durations!

See example: https://play.golang.org/p/WfFN0VHxoRY

logging/measure/model/apitrace/apitrace.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+// Package apitrace provides data model for binary API trace informations.
+package apitrace;
+
+message Trace {
+    message Entry {
+        uint64 index = 1;                       // Trace index
+        string msg_name = 2;                    // Message name (or any other string)
+        uint64 duration = 3;                    // Duration as integer
```
ondrej-fabry: You probably meant `... as nanoseconds`, the integer is obvious from definition.

logging/measure/model/apitrace/apitrace.proto
```diff
@@ -0,0 +1,20 @@
...
+// Package apitrace provides data model for binary API trace informations.
+package apitrace;
+
+message Trace {
+    message Entry {
+        uint64 index = 1;                       // Trace index. Every message is indexed starting with 1 in order they
+                                                // called in in the VPP
+        string msg_name = 2;                    // Message name (or any other descriptive string)
+        uint64 duration = 3;                    // Duration of message call in nanoseconds
+    }
```
ondrej-fabry: Entries should definitely contain start time as well so we can later filter them and see how close they are to each other.

logging/measure/model/apitrace/apitrace.proto
```diff
@@ -0,0 +1,20 @@
...
+    repeated Entry traced_entries = 1;
+    message Average {
+        string msg_name = 1;                    // Message name, unique in this list.
+        uint64 average_time = 2;                // Average time duration per message in nanoseconds
```
ondrej-fabry: To be consistent with field in `Entry`, this should be named `average_duration`.

logging/measure/tracer.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// Single time entry
+type entry struct {
+	index      uint64
+	name       string
+	loggedTime time.Duration
+}
+
+func (t *tracer) LogTime(entity string, start time.Time) {
```
ondrej-fabry: This is good as very raw form of adding entries, but there should be an API creating entry which would store current time and then return it and later be able to call Log/Record for the entry to actually calculate duration and store the entry in `timedb`. 

This allows users to not have to take care of their start time. And more importantly this will store the index at the point of entry creation and not when it gets logged. That way all indexes for entries will be ordered chronologically just like they started.

logging/measure/tracer.go
```diff
@@ -0,0 +1,125 @@
...
+
+// Inner structure handling database and measure results
+type tracer struct {
+	sync.Mutex
+
+	name string
+	log  logging.Logger
+	// Entry index, used in database as key and increased after every entry. Never resets since the tracer object is
+	// created or the database is cleared
+	index uint64
```
ondrej-fabry: This should be named `nextIndex` so it's obvious that it does not represent any actual valid index (current value not used in any entry).

logging/measure/tracer.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+
+	"github.com/ligato/cn-infra/logging"
+	"github.com/ligato/cn-infra/logging/measure/model/apitrace"
+)
+
+// Tracer allows to measure, store and list measured time entries.
+type Tracer interface {
+	// LogTime puts measured time to the table and resets the time.
+	LogTime(entity string, start time.Time)
```
ondrej-fabry: The names `entity`, `name` and `msg_name` seem to be mixed.

logging/measure/model/apitrace/apitrace.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+
+// Package apitrace provides data model for binary API trace informations.
+package apitrace;
+
+message Trace {
+    message Entry {
+        uint64 index = 1;                       // Trace index. Every message is indexed starting with 1 in order they
+                                                // called in in the VPP
+        string msg_name = 2;                    // Message name (or any other descriptive string)
```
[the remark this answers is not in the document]
VladoLavor: If we don't want to store all the data from the call (which I think is a huge overkill), I don't see a point in adding custom fields. What we should store in it by default?

PR #342, opened 2018-09-25
basic rest security using tokens

VladoLavor:
- authentication enabled/disabled in the config file
- users are loaded from the config file, except admin which is always present (otherwise rest may get inaccessible if no users are loaded since registration is not available yet)
- permission groups can be added to users with URLs specified (again, except admin group which is always present and has access to everything)
- login generates token, log out invalidates it. Token may also expire if time is set in the config file


rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+// Authenticator keeps information about users, permission groups and tokens and processes it
+type authenticator struct {
+	log logging.Logger
+
+	// Router instance automatically registers login/logout REST API handlers if authentication is enabled
+	router    *mux.Router
+	formatter *render.Render
+
+	// User database keeps all known users with permissions and hashed password. Users are loaded from
+	// HTTP config file // todo add other options
```
ondrej-fabry: _TODOs_ should always be defined on separate line with all caps, otherwise it's not gonna get noticed very easily.

rpc/rest/security/model/http-security/httpsecurity.proto
```diff
@@ -0,0 +1,15 @@
...
+// Package http_security provides basic definition of user and permission group
+package http_security;
+
+message User {
+    string name = 1;
+    string password = 2;
+    repeated string permissions = 3;    // Permission name, user is allowed for more permission groups
+}
+
+message PermissionGroup {
```
ondrej-fabry: I think `PermissionGroup` and `Permission` should be separate messages, otherwise it's not gonna be easy to extend this functionality.

First thing that comes to mind is allowing only some of the HTTP methods (GET, POST..) to be allowed for single URL or defining URL prefix as a whole in single permission, etc.

Keep `name` field in the `PermissionGroup` and replace `urls` array with array of `Permission` messages which will have: `url` and `allowed_methods` (string array).

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+
+// NewAuthenticator prepares new instance of authenticator.
+func NewAuthenticator(router *mux.Router, expTime int, users []httpsecurity.User, cost int, log logging.Logger) AuthenticatorAPI {
+	a := &authenticator{
+		router: router,
+		log:    log,
+		formatter: render.New(render.Options{
+			IndentJSON: true,
+		}),
+		userDb:  make([]*httpsecurity.User, 0),
```
ondrej-fabry: Not needed.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+		Permissions: []string{admin},
+	})
+
+	go func() {
+		// Process users in go routine, since hashing may take some time
+		for _, user := range users {
+			// Password hash cost is currently set via config file. Values allowed by bcrypt are in range 4-31. The min
+			// value will be set if cost is not defined in config file. Keep in mind that high cost values (>10) can
+			// take a lot of time to process.
+			hash, err := bcrypt.GenerateFromPassword([]byte(user.Password), cost)
```
ondrej-fabry: This is not correct! Password can never be stored as plaintext! Only hashed password are loaded from config file.

rpc/rest/security/model/http-security/httpsecurity.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+
+// Package http_security provides basic definition of user and permission group
+package http_security;
+
+message User {
+    string name = 1;
+    string password = 2;
```
ondrej-fabry: It should be explicit that the password is hashed, please name it `hashed_password`.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+				Name:        user.Name,
+				Password:    string(hash),
+				Permissions: user.Permissions,
+			})
+			a.log.Debugf("Registered user %s, permissions: %v", user.Name, user.Permissions)
+		}
+	}()
+
+	// Admin-group, available by default and always enabled for all URLs
+	a.AddPermissionGroup(&httpsecurity.PermissionGroup{
```
ondrej-fabry: I think there is no need to use exported method for adding default `admin` group, you can just directly initialize the `groupDb` with it when creating `authenticator`.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+
+	// Router instance automatically registers login/logout REST API handlers if authentication is enabled
+	router    *mux.Router
+	formatter *render.Render
+
+	// User database keeps all known users with permissions and hashed password. Users are loaded from
+	// HTTP config file // todo add other options
+	userDb []*httpsecurity.User
+	// Permission database is a map of permissions and bound URLs
+	groupDb []*httpsecurity.PermissionGroup
```
ondrej-fabry: This should probably be map keyed by group name for faster and simpler lookups without extra loops.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+			// value will be set if cost is not defined in config file. Keep in mind that high cost values (>10) can
+			// take a lot of time to process.
+			hash, err := bcrypt.GenerateFromPassword([]byte(user.Password), cost)
+			if err != nil {
+				a.log.Errorf("failed to hash password for user %s: %v", user.Name, err)
+				continue
+			}
+
+			a.userDb = append(a.userDb, &httpsecurity.User{
+				Name:        user.Name,
```
ondrej-fabry: How do you deal with name collisions here?

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+type authenticator struct {
+	log logging.Logger
+
+	// Router instance automatically registers login/logout REST API handlers if authentication is enabled
+	router    *mux.Router
+	formatter *render.Render
+
+	// User database keeps all known users with permissions and hashed password. Users are loaded from
+	// HTTP config file // todo add other options
+	userDb []*httpsecurity.User
```
ondrej-fabry: Users and permissions should not be stored as fields directly in `authenticator`. Authenticator should access them through interface, so we can use other implementations with different storage methods. 

Create interface for retrieving these, something like:

```go
type AuthStore interface {
    GetUser(name string) (*User, error)
    GetAllUsers() ([]*User, error)
    AddUser(user *User) error
    // permissions, etc...
}
```

and then create default implementation which will store them in memory, similarity like you did here and make it possible to set different storage through authenticator constructor.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+	// Add admin-user, enabled by default, always has access to every URL
+	users = append(users, httpsecurity.User{
+		Name:        admin,
+		Password:    "ligato123",
+		Permissions: []string{admin},
+	})
+
+	go func() {
+		// Process users in go routine, since hashing may take some time
+		for _, user := range users {
```
ondrej-fabry: You should check and reject any user named `admin`.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+
+// Validates credentials, returns name and error code/message if invalid
+func (a *authenticator) validateCredentials(req *http.Request) (string, int, error) {
+	name, pass, ok := req.BasicAuth()
+	if !ok {
+		return name, http.StatusInternalServerError, errors.Errorf("500 internal server error: unable to process authentication data")
+	}
+	user := a.getUser(name)
+	if user == nil {
+		return name, http.StatusUnauthorized, errors.Errorf("401 unauthorized: user name does not exist")
```
ondrej-fabry: For security reasons, the error message should be exactly same for invalid user and invalid password. Otherwise, an attacker will be able to do brute force attack more easily.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+	name, _, ok := req.BasicAuth()
+	if !ok {
+		a.formatter.Text(w, http.StatusInternalServerError, "500 internal server error: unable to process authentication data")
+		return
+	}
+	delete(a.tokenDb, name)
+}
+
+// Validates credentials, returns name and error code/message if invalid
+func (a *authenticator) validateCredentials(req *http.Request) (string, int, error) {
```
ondrej-fabry: If you are gonna extract this function than it should just check user/password combination and not care about request at all, otherwise it doesn't really help to extract it, because it's coupled with http stuff.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/pkg/errors"
+	"github.com/unrolled/render"
+	"golang.org/x/crypto/bcrypt"
+)
+
+const (
+	// Helps to obtain authorization header matching the field in a request
+	authHeaderStr = "authorization"
+	// Used to sign the token // todo maybe use something more secure
+	secret = "secret"
```
ondrej-fabry: This needs to be loaded from config or env, otherwise anyone can simply guess or read it from the source and easily manipulate token data.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+		return
+	}
+	var claims jwt.MapClaims = make(map[string]interface{})
+	if a.expTime != 0 {
+		claims["exp"] = time.Now().Add(time.Minute * time.Duration(a.expTime)).Unix()
+	}
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	tokenString, err := token.SignedString([]byte(secret))
+	if err != nil {
+		errStr := fmt.Sprintf("500 internal server error: faield to sign token: %v", err)
```
ondrej-fabry: There is a typo and also such message should probably be printed in the log as well.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+	groupDb []*httpsecurity.PermissionGroup
+	// Token database keeps information of actual token and its owner.
+	tokenDb map[string]string
+
+	// Token claims
+	expTime int
+}
+
+// NewAuthenticator prepares new instance of authenticator.
+func NewAuthenticator(router *mux.Router, expTime int, users []httpsecurity.User, cost int, log logging.Logger) AuthenticatorAPI {
```
ondrej-fabry: The expiration time should be as `time.Duration` type here.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+	if !ok {
+		a.formatter.Text(w, http.StatusInternalServerError, "500 internal server error: unable to process authentication data")
+		return
+	}
+	delete(a.tokenDb, name)
+}
+
+// Validates credentials, returns name and error code/message if invalid
+func (a *authenticator) validateCredentials(req *http.Request) (string, int, error) {
+	name, pass, ok := req.BasicAuth()
```
ondrej-fabry: The name/password should not come from `BasicAuth()`, that is for different type of security and if you look into implementation for it, you will find it will actually clash with this authorization.

Credentials should be given via JSON, hence the name *REST security*. Parse request body into structure to retrieve user/password input.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	tokenString, err := token.SignedString([]byte(secret))
+	if err != nil {
+		errStr := fmt.Sprintf("500 internal server error: faield to sign token: %v", err)
+		a.formatter.Text(w, http.StatusInternalServerError, errStr)
+		return
+	}
+	a.tokenDb[name] = tokenString
+	a.formatter.Text(w, http.StatusOK, tokenString)
```
ondrej-fabry: We are doing REST security, why are we returning plan text here?

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+// Validates credentials and provides new token
+func (a *authenticator) createTokenEndpoint(w http.ResponseWriter, req *http.Request) {
+	name, errCode, err := a.validateCredentials(req)
+	if err != nil {
+		a.formatter.Text(w, errCode, err.Error())
+		return
+	}
+	var claims jwt.MapClaims = make(map[string]interface{})
+	if a.expTime != 0 {
+		claims["exp"] = time.Now().Add(time.Minute * time.Duration(a.expTime)).Unix()
```
ondrej-fabry: You should probably use `ExpiresAt` field from [StandardClaims](https://godoc.org/github.com/dgrijalva/jwt-go#example-NewWithClaims--StandardClaims) for this. 

Also, you should store username here as well.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+// Validate the request
+func (a *authenticator) Validate(provider http.HandlerFunc) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		authHeader := req.Header.Get(authHeaderStr)
+		if authHeader == "" {
+			a.formatter.Text(w, http.StatusUnauthorized, "401 Unauthorized: authorization header required")
+			return
+		}
+		bearerToken := strings.Split(authHeader, " ")
+		if len(bearerToken) != 2 {
```
ondrej-fabry: Please check if the first part is equal to `Bearer`, otherwise the header entry is not valid.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+			a.formatter.Text(w, http.StatusUnauthorized, "401 Unauthorized: invalid authorization token")
+			return
+		}
+		token, err := jwt.Parse(bearerToken[1], func(token *jwt.Token) (interface{}, error) {
+			if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
```
ondrej-fabry: Please use [GetSigningMethod](https://godoc.org/github.com/dgrijalva/jwt-go#GetSigningMethod) to validate this.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+
+// Validates credentials and provides new token
+func (a *authenticator) createTokenEndpoint(w http.ResponseWriter, req *http.Request) {
+	name, errCode, err := a.validateCredentials(req)
+	if err != nil {
+		a.formatter.Text(w, errCode, err.Error())
+		return
+	}
+	var claims jwt.MapClaims = make(map[string]interface{})
+	if a.expTime != 0 {
```
ondrej-fabry: The expiration time can never be indefinite, the token would be valid forever, it always has to expire. If expiration time is set to `0` use some sane default as one hour.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,304 @@
...
+				continue
+			}
+		}
+		a.log.Debugf("added HTTP permission group %s", newPermissionGroup.Name)
+		a.groupDb = append(a.groupDb, newPermissionGroup)
+	}
+}
+
+// Validate the request
+func (a *authenticator) Validate(provider http.HandlerFunc) http.HandlerFunc {
```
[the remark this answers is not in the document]
VladoLavor: I see... I guess I removed it accidentally, which is odd because token expiration worked

rpc/rest/security/rest_security.go
VladoLavor: plain text is easier to copy because you can just take all the output

rpc/rest/security/rest_security.go
VladoLavor: will be fixed in following update

rpc/rest/security/rest_security.go
VladoLavor: Ok, but I'm just curious how it clashes with tokens.  Basic authorization reads name/password from authorization header, not sure what is wrong with it.

rpc/rest/security/rest_security.go
VladoLavor: I actually hate this approach, because when I try to remember some credentials, I never know what I put wrong :-)

rpc/rest/security/rest_security.go
ondrej-fabry: It will clash if the basic authorization currently supported in rest (http) plugin is enabled:

https://github.com/ligato/cn-infra/blob/master/rpc/rest/auth.go#L25

ondrej-fabry: Also, you are taking token from `Authorization` header:
https://github.com/ligato/cn-infra/pull/342/files/42f1af81d8ad534da5cf146b2d40e81cada60ae0#diff-2d725d9a6ca3d13bf2f95cae0e1139bfR149

`BasicAuth` takes value from `Authorization` header as well:
https://golang.org/src/net/http/request.go#L872

rpc/rest/security/rest_security.go
ondrej-fabry: Double-edged sword I know..

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,415 @@
...
+	authenticate = "/authenticate"
+	// Cookie name identifier
+	cookieName = "ligato-rest"
+)
+
+// Default value to sign the token, if not provided from config file
+var signature = "secret"
+
+// Default expiration time for token/cookie
+var expTime time.Duration = 3600000000000 // 1 Hour
```
ondrej-fabry: What purpose does this serve instead of having shorter declaration which does not even nee to be commented:

```go
var expTime = time.Hour
```

rpc/rest/security/model/http-security/httpsecurity.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+// Package http_security provides basic definition of user and permission group
+package http_security;
```
ondrej-fabry: After yesterday's call, we know that this will need to be more generic and serve other uses than just HTTP. We should rename it to `AccessSecurity`.

rpc/rest/security/model/http-security/httpsecurity.proto
```diff
@@ -0,0 +1,19 @@
...
+message User {
+    string name = 1;
+    string password_hash = 2;
+    repeated string permissions = 3;            // Permission name, user is allowed for more permission groups
+}
+
+message PermissionGroup {
+    string name = 1;
+    message Permissions {
+        string url = 1;                         // URL enabled for the permission
```
ondrej-fabry: This might also need to be little more generic instead of specifics like: *URL* or *method* so it can more easily fit other types, like the GRPC. Try to think about some alternatives that would fit both.

rpc/rest/security/auth_store.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/ligato/cn-infra/rpc/rest/security/model/http-security"
+)
+
+// StorageType defines user storage type
+type StorageType string
```
ondrej-fabry: This shouldn't be needed other implementations will be implemented without having to edit this package, this one provides only the default in-memory version.

rpc/rest/security/auth_store.go
```diff
@@ -0,0 +1,135 @@
...
+	lastLogout time.Time
+}
+
+// defaultAuthStorage is default implementation of AuthStore
+type defaultAuthStorage struct {
+	db []*User
+}
+
+// CreateAuthStore builds new storage type of provided type
+func CreateAuthStore(storageType StorageType) AuthStore {
```
ondrej-fabry: ..then, this should explicitly have "memory" or "mem" in the name so it's obvious what this implementation is for.

rpc/rest/security/auth_store.go
```diff
@@ -0,0 +1,135 @@
...
+// AuthStore is common interface to access user database/permissions
+type AuthStore interface {
+	// AddUser adds new user with name, password and permission groups. Password should be already hashed.
+	AddUser(name, passwordHash string, permissions []string) error
+	// GetUser returns user data according to name, or nil of not found
+	GetUser(name string) (*User, error)
+	// SetLoginTime writes last login time for specific user
+	SetLoginTime(name string)
+	// GetLoginTime returns las login time or error if user does not exist
+	GetLoginTime(name string) (time.Time, error)
```
ondrej-fabry: Do we need this, since we already have both login and logout times inside `User`?

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,415 @@
...
+	// permission group enabled has access to that set of keys. PGs with duplicated names are skipped.
+	AddPermissionGroup(group ...*httpsecurity.PermissionGroup)
+
+	// Validate serves as middleware used while registering new HTTP handler. For every request, token
+	// and permission group is validated.
+	Validate(provider http.HandlerFunc) http.HandlerFunc
+}
+
+// Context defines fields required to instantiate authenticator
+type Context struct {
```
ondrej-fabry: The word **context** and also use of such package is generally used for request scoped (time based) contexts, which this isn't at all. This is either *options*, *config* or *settings*, since those represents something persistent.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,415 @@
...
+
+// NewAuthenticator prepares new instance of authenticator.
+func NewAuthenticator(router *mux.Router, ctx *Context, log logging.Logger) AuthenticatorAPI {
+	a := &authenticator{
+		router: router,
+		log:    log,
+		formatter: render.New(render.Options{
+			IndentJSON: true,
+		}),
+		userDb:  CreateAuthStore(ctx.StorageType),
```
ondrej-fabry: The interface `AuthStore` should come as parameter to this constructor to allow users of this package use their own implementations if they want.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,415 @@
...
+
+	// Set token signature
+	signature = ctx.Signature
+	if a.expTime == 0 {
+		a.expTime = expTime
+		a.log.Debugf("Token expiration time claim not set, defaulting to 1 hour")
+	}
+
+	// Add admin-user, enabled by default, always has access to every URL
+	hash, err := bcrypt.GenerateFromPassword([]byte("ligato123"), ctx.Cost)
```
ondrej-fabry: This might potentially be expensive operation. Could you do it in goroutine? but keep in mind that needs to be concurrently safe, perhaps at the auth store level.

rpc/rest/config.go
```diff
@@ -88,6 +89,21 @@ type Config struct {
 	// ClientCerts is a slice of the root certificate authorities
 	// that servers uses to verify a client certificate
 	ClientCerts []string `json:"client-cert-files"`
+
+	// EnableTokenAuth enables token authorization for HTTP requests
+	EnableTokenAuth bool `json:"enable-token-auth"`
```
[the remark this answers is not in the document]
VladoLavor: Of course we want to have single security implementation for all RPC plugins, but this PR adds security for REST plugin only, so I'll keep the configuration inside it for now.

rpc/rest/plugin_api_rest.go
```diff
@@ -36,6 +37,9 @@ type HTTPHandlers interface {
 	// RegisterHTTPHandler propagates to Gorilla mux
 	RegisterHTTPHandler(path string, provider HandlerProvider, methods ...string) *mux.Route
 
+	// RegisterPermissionGroup registers new permission groups for users
+	RegisterPermissionGroup(group ...*httpsecurity.PermissionGroup)
```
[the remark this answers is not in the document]
VladoLavor: Same here, middleware required for REST was specified. Possibility to add custom middleware can be done together with a new generic security plugin as you mentioned above.

rpc/rest/security/model/http-security/httpsecurity.proto
VladoLavor: If we want to use a single model to define permission groups for all RPC types, I think it would be better to define a separate set of permission field for each of them, instead of thinking about some generic fields which may become confusing.

rpc/rest/security/model/access-security/accesssecurity.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+// Package http_security provides basic definition of user and permission group
+package http_security;
```
ondrej-fabry: Forgot to change it here?

rpc/rest/plugin_impl_rest.go
```diff
@@ -70,6 +74,18 @@ func (p *Plugin) Init() (err error) {
 		IndentJSON: true,
 	})
 
+	// Enable authentication if defined by config
+	if p.EnableTokenAuth {
+		p.Log.Info("Token authentication for HTTP enabled")
+		p.auth = security.NewAuthenticator(p.mx, &security.Settings{
+			// Since there is currently only one storage type, set it directly
```
ondrej-fabry: Obsolete comment.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+}
+
+// Validate the request
+func (a *authenticator) Validate(provider http.HandlerFunc) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		// Token may be accessed via cookie, or from authentication header
+		var tokenString string
+		// Try to read the cookie first
+		cookie, err := req.Cookie(cookieName)
```
ondrej-fabry: This should be as fallback not as first choice.

ondrej-fabry: And perhaps just extract this retrieval of token from request to separate function.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,422 @@
...
+	if req.Method == http.MethodGet {
+		r := render.New(render.Options{
+			Directory:  "templates",
+			Asset:      Asset,
+			AssetNames: AssetNames,
+		})
+		r.HTML(w, http.StatusOK, "login", nil)
+	}
+	// POST decodes provided credentials
+	if req.Method == http.MethodPost {
```
ondrej-fabry: Why is this not in else case? It needs to be evaluated only if first one is false.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,422 @@
...
+		// Returns token string.
+		a.formatter.Text(w, http.StatusOK, token)
+	}
+}
+
+// Authentication handler verifies credentials from login page (GET) and writes cookie with token
+func (a *authenticator) authenticationHandler(w http.ResponseWriter, req *http.Request) {
+	// Read name and password from the form (if accessed from browser)
+	credentials := &credentials{}
+	credentials.Username = req.FormValue("name")
```
ondrej-fabry: Initialize all in single statement.

rpc/rest/security/rest_security.go
VladoLavor: I put the cookie check first because it can return only "not found" error, but since we don't know whether the user is accessing URL via browser or curl, we can't return the error. 
Also, I think it does not matter what we try first.

But I put the code to the separate method.

rpc/rest/security/rest_security.go
VladoLavor: ok but honestly, do we really save something here?

rpc/rest/security/rest_security.go
ondrej-fabry: > we don't know whether the user is accessing URL via browser or curl

This is simply not true. Every HTTP client will sends information about "user agent" used inside the HTTP header `User-Agent`. For curl it will print `curl/X.XX.X`. You can retrieve it using `req.Header.Get("User-Agent")`. 

However such check should not be needed here! The most common way to use JWT is with Authorization header and it should always come first. What if cookies are stale and you are sending correct info through Authorization header? Then until you delete cookies it won't work. 

Authorization header should always be checked first as the main source of thruth. Cookies are just some crutch, they are not meant to be used for security related things at all, but they can serve good enough fallback here so we cant easily use this in the browsers without coding Javascript clients.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,449 @@
...
+	"github.com/ligato/cn-infra/logging"
+	access "github.com/ligato/cn-infra/rpc/rest/security/model/access-security"
+	"github.com/pkg/errors"
+	"github.com/unrolled/render"
+	"golang.org/x/crypto/bcrypt"
+)
+
+const (
+	// Helps to obtain authorization header matching the field in a request
+	authLabel = "authorization"
```
ondrej-fabry: This should start with capital letter just like it is defined in any specification. Also it's not a label, it's HTTP request header.

rpc/rest/security/auth_store.go
```diff
@@ -0,0 +1,133 @@
...
+	// AddUser adds new user with name, password and permission groups. Password should be already hashed.
+	AddUser(name, passwordHash string, permissions []string) error
+	// GetUser returns user data according to name, or nil of not found
+	GetUser(name string) (*User, error)
+	// SetLoginTime writes last login time for specific user
+	SetLoginTime(name string)
+	// SetLoginTime writes last logout time for specific user
+	SetLogoutTime(name string)
+	// IsLoggedOut uses login/logout timestamps to evaluate whether the user was logged out
+	IsLoggedOut(name string) (bool, error)
```
ondrej-fabry: Why is this needed? The user does not have to logout.. we don't really keep any session information so we can't say this with certainty. Token will expire and that's the logout.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,449 @@
...
+		var ok bool
+		if userName, ok = v["aud"].(string); !ok {
+			return fmt.Errorf("failed to validate token claims audience")
+		}
+	case jwt.StandardClaims:
+		userName = v.Audience
+	default:
+		return fmt.Errorf("failed to validate token claims")
+	}
+	loggedOut, err := a.userDb.IsLoggedOut(userName)
```
ondrej-fabry: The token will log user out by expiring.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	// Process users in go routine, since hashing may take some time
+	go func() {
+		// Add admin-user, enabled by default, always has access to every URL
+		hash, err := bcrypt.GenerateFromPassword([]byte("ligato123"), ctx.Cost)
+		if err != nil {
+			a.log.Errorf("failed to hash password for admin: %v", err)
+		}
+		a.userDb.AddUser(admin, string(hash), []string{admin})
```
ondrej-fabry: This could fail too you know, you don't know what type of auth store is used.

rpc/rest/security/auth_store.go
ondrej-fabry: https://medium.com/devgorilla/how-to-log-out-when-using-jwt-a8c7823e8a6

ondrej-fabry: I'm not saying anything against logout feature and also nothing about storing tokens. It's just that the logic how the logout is made now still uses some stored state, which will actually not be stored if application restarts. In the article they imply that the token should be replaced with one with very short expiration time, or even expired. Which after replacing it in the cookie will make it unable to login even if application restarts, because the token will be invalid.

rpc/rest/security/auth_store.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (ds *defaultAuthDB) AddUser(name, passwordHash string, permissions []string) error {
+	ds.Lock()
+	defer ds.Unlock()
+
+	// Verify user does not exist yet
+	var exists bool
+	for _, userData := range ds.db {
+		if userData.Name == name {
+			exists = true
```
ondrej-fabry: Why not return here?

rpc/rest/security/model/access-security/accesssecurity.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+// Package http_security provides basic definition of user and permission group
```
ondrej-fabry: Forgot to change comment here.

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,451 @@
...
+	signature = ctx.Signature
+	if a.expTime == 0 {
+		a.expTime = defaultExpTime
+		a.log.Debugf("Token expiration time claim not set, defaulting to 1 hour")
+	}
+
+	// Process users in go routine, since hashing may take some time
+	go func() {
+		// Add admin-user, enabled by default, always has access to every URL
+		hash, err := bcrypt.GenerateFromPassword([]byte("ligato123"), ctx.Cost)
```
ondrej-fabry: Please don't store plaintext here, it's not safe and it's is really unnecessary to generate hash from a known password every time. Just store hashed password. Only generating password needs cost, comparing can be done with any cost since the cost is retrievable from the hashed password, hence: https://godoc.org/golang.org/x/crypto/bcrypt#Cost

rpc/rest/security/rest_security.go
```diff
@@ -0,0 +1,451 @@
...
+	// Writes cookie with token.
+	http.SetCookie(w, &http.Cookie{
+		Name:   cookieName,
+		Path:   "/",
+		MaxAge: int(a.expTime.Seconds()),
+		Value:  token,
+		Secure: false,
+	})
+	// Automatically move to index page.
+	target := "http://" + req.Host + "/"
```
ondrej-fabry: Why is this redirecting to `http`, what if HTTPS is used? It should be enough to use `"/"` in redirect.

rpc/rest/security/auth_store.go
VladoLavor: Replace a cookie with short expiration time is one thing (and you can easily get around it copying old token and putting it back after log out), but we cannot invalidate tokens obtained as plain text on the client side. 

I think we should decide what is more important - to have a strict logout, or if we want login to survive application restart.

rpc/rest/security/rest_security.go
VladoLavor: the default admin password need not be safe (after all it's written in the readme), in the future, it will be possible to change it and then the hash has to be calculated anyway. But I agree, it is not needed to do it every time during startup.

PR #343, opened 2018-09-26
Kvscheduler revisited

milanlenco:
- KVDescriptor is now a structure
- Value is always proto.Message
- internal stuff moved under internal subdirectory
- log shortened, beautified
- transaction processing simplified, some bugs were also fixed
- added callback `RetriableFailure` to the descriptor that will allow to disable retries for errors that are guaranteed to never be fixed by just repeating the same operation with the same data
- added new feature called "HalfwayResync" (proposals for different names are welcomed) to sync between scheduler and SB (as opposed to FullResync which also performs sync between NB and scheduler)
- simplified and cleaned up API a bit
- added REST API for dumping of kv-pairs under a given descriptor (of SB or scheduler's internal state)

I plan to add some more UTs, but before I would rebase Linux plugin and re-test if it still works.


ondrej-fabry:
>     * added new feature called `HalfwayResync`

What about `DownstreamResync` to make it more clear that it's about resync southband.

kvscheduler/api/errors.go
```diff
@@ -62,3 +73,16 @@ func ErrInvalidMetadataType(key string) error {
 	}
 	return fmt.Errorf("metadata has invalid type for key: %s", key)
 }
+
+// NonRetriableIfInTheList returns callback suitable for KVDescriptor.RetriableFailure
+// based on an exhaustive list of all possible non-retriable errors.
+func NonRetriableIfInTheList(errorList []error) func(err error) bool {
+	return func(err error) bool {
+		for _, nonRetriableError := range errorList {
+			if err.Error() == nonRetriableError.Error() {
```
ondrej-fabry: Not sure what is the intended use case for this function, since it's used internally only and the loop over the errors can be done there. Also you should compare the types directly, not the string output from `Error`.

milanlenco: It is helper for descriptors to generate callback for:
`RetriableFailure func(err error) bool`

But perhaps instead of the callback it could be just a list of non-retriable errors. Would it be sufficient for any descriptor?

ondrej-fabry: I would keep the way it is now and let each descriptor handle thier errors inside `RetriableFailure` function defined in Descriptor on their own and just get rid of this helper method, don't be afraid to be verbose here. Better to keep the logic of deciding which error is retriable inside the function in Descriptor.

kvscheduler/api/kv_descriptor_api.go
```diff
@@ -90,76 +91,109 @@ func (vo ValueOrigin) String() string {
...
 	// preventing the existence of this value.
-	Update(key string, value Value, metadata Metadata) error
+	Update func(key string, value proto.Message, metadata Metadata) error
+
+	// RetriableFailure tells scheduler if the given error, returned by one
+	// of Add/Delete/Modify/Update handlers, will always be returned for the
+	// the same value (non-retriable) or if the value can be theoretically
+	// fixed merely by repeating the operation.
+	// If the callback is not defined, every error will be considered retriable.
+	RetriableFailure func(err error) bool
```
ondrej-fabry: Probably better to keep all boolean-returning functions with `Is..` prefix.

milanlenco:
> What about `DownstreamResync` to make it more clear that it's about resync southband.

Sounds good. How would we call `the full resync` then (within the scheduler)? `FullResync` or just `Resync`?

ondrej-fabry:
> Sounds good. How would we call `the full resync` then (within the scheduler)? `FullResync` or just `Resync`?

Probably keep it it more explicit with `FullResync` to make it standout from the other type.

milanlenco:
OK, `FullResync` it is.
The PR is ready for merge then. I'm preparing some more UTs, but they can go in another PR, so that datasync refactor is not blocked.

PR #350, opened 2018-10-10
filesystem as key-value data store

VladoLavor:
Supported features:
* files/directories to watch can be defined in config file (--filesystem-config=(path))
* read configuration is applied during resync (same way as other kv stores)
* changes in config files during runtime are propagated to plugin, and appropriate events are generated


Makefile
```diff
@@ -151,6 +151,7 @@ MDLINKCHECK := $(shell command -v markdown-link-check 2> /dev/null)
 # Get link check tool
 get-linkcheck:
 ifndef MDLINKCHECK
+	sudo apt-get update
```
ondrej-fabry: This should go after the next log line and ideally combined together with install, like: `sudo apt-get update && sudo apt-get install npm`.

db/keyval/filesystem/client.go
```diff
@@ -0,0 +1,355 @@
...
+type Client struct {
+	sync.Mutex
+	log logging.Logger
+
+	// A list of filesystem paths. It may be specific file, or the whole directory. In case a path is a directory,
+	// all files within are processed.
+	paths []string
+
+	// Internal database mirrors changes in file system. Since the configuration can be only read, it is up to client
+	// to handle difference between configuration revisions. Evert database entry consists from three values; path
```
ondrej-fabry: `Evert` typo.

db/keyval/filesystem/client.go
```diff
@@ -0,0 +1,355 @@
...
+	// of a file where the configuration is written, data key and data value.
```
ondrej-fabry: Perhaps these 3 values could be listed with bullet points:
```
- path of a source file
- data key
- data value
```

db/keyval/filesystem/reader/reader.go
```diff
@@ -0,0 +1,189 @@
...
+	// ProcessFile returns File type object. Provided path must be a file
+	ProcessFile(path string) (File, error)
+	// ProcessFilesInDir returns list of File type objects. Provided path must be a directory
+	ProcessFilesInDir(path string) ([]File, error)
+	// IsValid validates interface
+	IsValid(ev fsnotify.Event) (bool, error)
+}
+
+// Reader implements file system API
+type Reader struct{}
```
ondrej-fabry: It seems to me that this reader API implementation should actually be called `JsonReader`, because it's tightly coupled with json encoding/decoding at the moment.

db/keyval/filesystem/client.go
```diff
@@ -0,0 +1,355 @@
...
+		db:          NewDbClient(),
+		r:           rd,
+		agentPrefix: prefix,
+		log:         log,
+	}
+
+	// Do the initial read and store everything to internal database
+	for _, path := range paths {
+		// Register to watcher
+		watcher.Add(path)
```
ondrej-fabry: Error is being ignored here.

db/keyval/filesystem/client.go
```diff
@@ -0,0 +1,355 @@
...
+// GetValue returns a value for given key
+func (c *Client) GetValue(key string) (data []byte, found bool, revision int64, err error) {
+	data, found = c.db.GetDataForKey(c.agentPrefix + key)
+	return
+}
+
+// ListValues returns a list of values for given prefix
+func (c *Client) ListValues(prefix string) (keyval.BytesKeyValIterator, error) {
+	keyValues := c.db.GetValuesForPrefix(c.agentPrefix + prefix)
+	data := make([]*reader.FileEntry, 0)
```
ondrej-fabry: You know the final capacity here, you can initialize the slice with `make([]*reader.FileEntry, 0, len(keyValues))` to prevent needs for reallocating the underlying array.

db/keyval/filesystem/reader/reader.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	return true, nil
+}
+
+// ProcessFile reads file from filesystem and un-marshalls it to the required proto structure
+func (r *Reader) ProcessFile(path string) (File, error) {
+	dataSet := File{}
+	fileData, err := ioutil.ReadFile(path)
+	if err != nil {
+		return dataSet, fmt.Errorf("failed to read file from path %s: %v", path, err)
```
ondrej-fabry: These functions should rather return pointer to a `File`, so we can return nil when error occurs and avoid returning something when not needed.

db/keyval/filesystem/reader/reader.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		files = append(files, file)
+	}
+
+	return files, nil
+}
+
+// IsDirectory checks if provided path is a file or a directory
+func (r *Reader) IsDirectory(path string) (bool, error) {
+	file, err := os.Open(path)
```
ondrej-fabry: This is not required to check if certain path is directory. Please use [os.Stat(path)](https://golang.org/pkg/os/#Stat).

db/keyval/filesystem/client.go
```diff
@@ -0,0 +1,355 @@
...
+}
+
+// GetWatcher returns filesystem watcher
+func (c *Client) GetWatcher() *fsnotify.Watcher {
+	return c.watcher
+}
+
+// Represents data with both identifiers, file path and key.
+type keyedData struct {
+	path string
```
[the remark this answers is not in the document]
VladoLavor: Yes, it should care - imagine the case where you rename or move configuration file somewhere out of the scope of the paths defined in the filedb config.
But I want to remove this struct completely anyway.

db/keyval/filesystem/client.go
```diff
@@ -0,0 +1,355 @@
...
+	return c.paths
+}
+
+// GetPrefix returns current agent prefix
+func (c *Client) GetPrefix() string {
+	return c.agentPrefix
+}
+
+// GetDB returns filesystem database
+func (c *Client) GetDB() FilesSystemDB {
```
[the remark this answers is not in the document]
VladoLavor: test purposes, only temporary change

db/keyval/filesystem/client.go
```diff
@@ -0,0 +1,355 @@
...
+	return c.db
+}
+
+// GetWatcher returns filesystem watcher
+func (c *Client) GetWatcher() *fsnotify.Watcher {
```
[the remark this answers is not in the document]
VladoLavor: same for this method

db/keyval/filesystem/reader/reader.go
VladoLavor: API changed, it is not coupled with the reader implementation anymore.

db/keyval/filesystem/reader/reader.go
VladoLavor: removed

VladoLavor:
One thing: json.RawMessage is specific for the given decoder, and since the value field is dynamic it cannot be unmarshalled directly to []byte.

db/keyval/filedb/decoder/decoder_json.go
```diff
@@ -0,0 +1,88 @@
...
+		})
+	}
+	// Encode to type specific structure
+	jsonFile := &jsonFile{Data: jsonFileEntries}
+	return json.Marshal(jsonFile)
+}
+
+// Decode provided byte set of json file and returns set of file data entries
+func (jd *JSONDecoder) Decode(byteSet []byte) ([]*FileDataEntry, error) {
+	if byteSet == nil || len(byteSet) == 0 {
```
ondrej-fabry: It's enough to check `len(byteSet) == 0`, since it will evaluate to true for `nil`.

db/keyval/filedb/decoder/decoder_yaml.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/ghodss/yaml"
+	"strings"
+)
+
+// Extension supported by this decoder
+var yamlDcExtensions = []string{".yaml"}
```
ondrej-fabry: This could be defined as exported field in `YAMLDecoder` to allow customization.

db/keyval/filedb/plugin.go
```diff
@@ -0,0 +1,131 @@
...
+	// Read fileDB configuration file
+	var err error
+	p.config, err = p.getFileDBConfig()
+	if err != nil || p.disabled {
+		return err
+	}
+
+	// Register decoders
+	var decoders []decoder.API
+	decoders = append(decoders, decoder.NewJSONDecoder(), decoder.NewYAMLDecoder())
```
ondrej-fabry: This can be initialized directly, no need for `append`.

db/keyval/filedb/filesystem/filesystem.go
```diff
@@ -0,0 +1,180 @@
...
+	// GetFiles takes a list of file system paths an returns a list of individual file paths
+	GetFileNames(paths []string) ([]string, error)
+	// Watch given paths, call 'onEvent' function on every event or 'onClose' if watcher is closed
+	Watch(paths []string, onEvent func(event fsnotify.Event), onClose func()) error
+	// Close allows to terminate watcher from outside which calls 'onClose'
+	Close() error
+}
+
+// FsHandler is helper struct to manipulate with filesystem API
+type FsHandler struct {
```
ondrej-fabry: The `Fs` prefix sounds like unnecessary stutter when used: `filesystem.FsHandler`. I think `filesystem.Handler` would suffice here.

db/keyval/filedb/filesystem/filesystem.go
```diff
@@ -0,0 +1,180 @@
...
+
+// CreateFile is an implementation of the file system API interface
+func (fsh *FsHandler) CreateFile(file string) error {
+	parts := strings.Split(file, "/")
+	if len(parts) == 0 {
+		return errors.Errorf("failed to create file: invalid input %s", file)
+	}
+	// Create path at first if necessary
+	if len(parts) > 1 {
+		path := strings.Replace(file, parts[len(parts)-1], "", 1)
```
ondrej-fabry: This seems complicated. There are handful of utility functions for this in the [filepath package](https://godoc.org/path/filepath). I think [filepath.Split](https://godoc.org/path/filepath#Split) can be used here to get rid of this low-level operations.

db/keyval/filedb/filesystem/filesystem.go
```diff
@@ -0,0 +1,180 @@
...
+		return false
+	}
+	return true
+}
+
+// GetFileNames is an implementation of the file system API interface
+func (fsh *FsHandler) GetFileNames(paths []string) (files []string, err error) {
+	for _, path := range paths {
+		err := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
+			if info == nil || info.IsDir() {
```
ondrej-fabry: For clarity this should check `err != nil` instead of `info == nil` they are mutually exclusive. Also if we are not going to log errors here, at least comment should be added that some files might silently fail to read.

db/keyval/filedb/filesystem/filesystem.go
```diff
@@ -0,0 +1,180 @@
...
+				return nil
+			}
+			files = append(files, path)
+			return nil
+		})
+		if err != nil {
```
ondrej-fabry: I don't think this can actually happen, because doc for [filepath.Walk](https://godoc.org/path/filepath#Walk) mentions that:

> All errors that arise visiting files and directories are filtered by walkFn.

And since your `walkFn` only ever returns nil, it cannot happen.

db/keyval/filedb/filesystem/filesystem.go
```diff
@@ -0,0 +1,180 @@
...
+			err = errors.Errorf("failed to traverse through %s: %v", path, err)
+		}
+	}
+	return files, err
+}
+
+// Watch starts new filesystem notification watcher. All events from of json/yaml type files are passed to 'onEvent' function.
+// Function 'onClose' is called when event channel is closed.
+func (fsh *FsHandler) Watch(paths []string, onEvent func(event fsnotify.Event), onClose func()) error {
```
ondrej-fabry: Instead needing two callbacks, you could use single channel as parameter in which the events would be sent and could be closed on Close. Also it could also be buffered if needed.

db/keyval/filedb/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Configuration
+
+All files/directories used as a data store must be defined in configuration file. Location of the file
+can be defined either by the command line flag `filedb-config` or set via the `FILEDB_CONFIG`
+environment variable.
+
+## Data structure
+
+Plugin currently supports only JSON and YAML-formatted data. The format of the file is as follows for JSON:
```
ondrej-fabry: List supported formats using list to make it more obvious at first glance, something like:

### Supported formats:
* JSON (`*.json`)
* YAML (`*.yaml`)

db/keyval/filedb/client.go
```diff
@@ -0,0 +1,422 @@
...
+	// Internal database mirrors changes in file system. Since the configuration can be only read, it is up to client
+	// to handle difference between configuration revisions. Every database entry consists from three values:
+	//  - path (where the configuration is written)
+	//  - data key
+	//  - data value
+	// Note: database holds only configuration intended for agent with appropriate prefix
+	db database.FilesSystemDB
+
+	// Filesystem handler, provides methods to work with files/directories
+	fsh filesystem.API
```
ondrej-fabry: Fields in structs, especially ones with several fields used across multiple files, should contain more verbose name for better code readability. It took me a few scrolls to figure out some of these fields. Few characters long variable names should mostly be used for local variables.

`fsh` -> `fsHandler`
`stPath` -> `statusPath`
`stDc` -> `statusDecoder`

db/keyval/filedb/database/database.go
```diff
@@ -0,0 +1,162 @@
...
+		if ok {
+			if bytes.Compare(value.data, entry.Value) != 0 {
+				rev := value.rev + 1
+				fileData[entry.Key] = &dbEntry{entry.Value, rev}
+			}
+		} else {
+			fileData[entry.Key] = &dbEntry{entry.Value, initialRev}
+		}
+	} else {
+		fileData = make(map[string]*dbEntry)
```
ondrej-fabry: This can be initialized directly.

db/keyval/filedb/database/database.go
```diff
@@ -0,0 +1,162 @@
...
+
+	if entry == nil {
+		return
+	}
+
+	fileData, ok := c.db[path]
+	if ok {
+		value, ok := fileData[entry.Key]
+		if ok {
+			if bytes.Compare(value.data, entry.Value) != 0 {
```
ondrej-fabry: It seems this should rather use `bytes.Equal` since you don't can how many differences there are.

db/keyval/filedb/client.go
```diff
@@ -0,0 +1,422 @@
...
+	return &BrokerWatcher{
+		Client: c,
+		prefix: prefix,
+	}
+}
+
+// Put reads status file, add data to it and performs write
+func (c *Client) Put(key string, data []byte, opts ...datasync.PutOption) error {
+	// Read and decode status file
+	stFile, err := c.fsh.ReadFile(c.stPath)
```
ondrej-fabry: This will need optimization, imagine 100x Put.. every time file needs to be read from disk, decoded, looped through, encoded, written to disk. That is too much overhead. Some caching should be introduced at the filesystem handler to prevent unnecessary re-reading.

If not implemented in this PR, at least some TODO should be mentioned here.

db/keyval/filedb/client.go
```diff
@@ -0,0 +1,422 @@
...
+
+	// File system decoder API, grants access to methods needed for decoding files. Client expects
+	// the decoder to process files with specific extension
+	decoders []decoder.API
+
+	// A set of watchers for every key prefix.
+	watchers map[string]chan keyedData
+
+	// Prefix to recognize configuration for specific agent instance
+	agentPrefix string
```
ondrej-fabry: The client should not know anything about some agent prefix. It is supposed to be low level implementation which does not need to know about how the keys are being used.

The `BrokerWatcher` should be prepending the prefix for the API calls not the other way around.

Look at the implementation for the consul:

https://github.com/ligato/cn-infra/blob/master/db/keyval/consul/consul.go#L366-L436

db/keyval/filedb/client.go
VladoLavor: Since the database is in sync with the filesystem, data do not need to be actually read from disk.

db/keyval/filedb/client.go
```diff
@@ -0,0 +1,379 @@
...
+func (c *Client) watch(resp func(response keyval.BytesWatchResp), dataChan chan keyedData, closeChan chan string, key string) {
+	for {
+		select {
+		case keyedData, ok := <-dataChan:
+			if !ok {
+				return
+			}
+			if keyedData.Op == datasync.Delete {
+				resp(&keyedData.watchResp)
+			} else if !bytes.Equal(keyedData.PrevValue, keyedData.Value) {
```
TrueFurby: Seems like `||` (or) can be used here.

db/keyval/filedb/client.go
```diff
@@ -0,0 +1,379 @@
...
+			byteFile, err := c.fsHandler.ReadFile(filePath)
+			if err != nil {
+				return nil, errors.Errorf("failed to read file %s content: %v", filePath, err)
+			}
+			fileEntries, err := dc.Decode(byteFile)
+			file := &decoder.File{Path: filePath, Data: fileEntries}
+			if err != nil {
+				return nil, errors.Errorf("failed to decode file %s: %v", filePath, err)
+			}
+			files = append(files, file)
```
TrueFurby: We could loop of file's data here, instead of saving to local `files` and do another loops below.

ondrej-fabry:
I'll merge this for now so it does not go stale. However I'd like to get back to it, because I think there are some approaches that should be considered.

PR #352, opened 2018-10-23
process manager plugin

VladoLavor:

### DIFF
--- a/processmanager/options.go
+++ b/processmanager/options.go
@@ -21,7 +21,7 @@ var DefaultPlugin = *NewPlugin()
 func NewPlugin(opts ...Option) *Plugin {
 	p := &Plugin{}
 
-	p.PluginName = "pl-manager"
+	p.PluginName = "process-manager"
 
 	for _, o := range opts {
 		o(p)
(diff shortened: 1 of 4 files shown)


processmanager/plugin.go
```diff
@@ -0,0 +1,94 @@
...
+
+// GetProcess uses identifier (name or command) to find a desired process. Also returns boolean flag whether
+// the instance was found
+func (p *Plugin) GetProcess(identifier string) (ProcessAPI, bool) {
+	for _, process := range p.processes {
+		if process.name == identifier || process.cmd == identifier {
+			return process, true
+		}
+	}
+	return nil, false
```
TrueFurby: Second parameters seems useless here, we could simply check if the first return value is nil.

processmanager/options.go
```diff
@@ -0,0 +1,43 @@
...
+package processmanager
+
+// DefaultPlugin is a default instance of Plugin.
+var DefaultPlugin = *NewPlugin()
+
+// NewPlugin creates a new Plugin with the provided Options.
+func NewPlugin(opts ...Option) *Plugin {
+	p := &Plugin{}
+
+	p.PluginName = "pl-manager"
```
TrueFurby: `pl-manager` means what here?

PR #355, opened 2018-11-08
Merge master and use jsonpb in JSON Serializer

ondrej-fabry:


db/keyval/proto_serializer.go
```diff
@@ -45,13 +43,21 @@ func (sp *SerializerProto) Marshal(message proto.Message) ([]byte, error) {
...
+	return jsonpb.Unmarshal(bytes.NewBuffer(data), protoData)
 }
 
-// Marshal serializes proto message to the slice of bytes using json marshaller.
+// Marshal serializes proto message to the slice of bytes using
+// jsonpb marshaller to correctly marshal protobuf data.
 func (sj *SerializerJSON) Marshal(message proto.Message) ([]byte, error) {
-	return json.Marshal(message)
+	var buf bytes.Buffer
+	if err := new(jsonpb.Marshaler).Marshal(&buf, message); err != nil {
```
ondrej-fabry: We have to use jsonpb, because otherwise it won't work with oneof fields.

PR #356, opened 2018-11-12
Extended process manager

VladoLavor:
* plugin package renamed from processmanager -> process
* added process monitoring
* the status file reading for data like CPU, memory usage...
* added several new methods to plugin and process API
* automatic restarts
* support for process templates


process/process.go
```diff
@@ -0,0 +1,240 @@
...
+	// GetPid returns process ID, or zero if process instance does not exist
+	GetPid() int
+	// UpdateStatus updates and returns all current plugin-defined process state data
+	UpdateStatus(pid int) (*status.File, error)
+	// GetCommand returns process command
+	GetCommand() string
+	// GetArguments returns process arguments if set
+	GetArguments() []string
+	// GetStartTime returns time when the process was started
+	GetStartTime() int
```
ondrej-fabry: Why doesn't this return `time.Time`?

process/template/template.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/ligato/cn-infra/logging"
+	"github.com/ligato/cn-infra/process/template/model/process"
+	"github.com/pkg/errors"
+)
+
+// JSONExt - JSON extension
+const JSONExt = ".json"
+
+// DefaultMode defines default permission bits for template file
+const DefaultMode = os.FileMode(int(0777))
```
ondrej-fabry: You shouldn't need to convert [untyped constants](https://golang.org/ref/spec#Constants). It's enough to defined it as: `os.FileMode(0777)`.

VladoLavor:
Sure thing, I will add some examples

examples/process-manager-plugin/test-process/test-process.go
```diff
@@ -0,0 +1,63 @@
...
+func (ta *TestApp) run() {
+	log.Info("test application started")
+	if *maxUptime != 0 {
+		log.Infof("max uptime set to %d second(s)", *maxUptime)
+	}
+
+	ticker := time.NewTicker(1 * time.Second)
+	var uptime uint
+
+	func() {
```
ondrej-fabry: Why is this wrapped in a function? You could replace `return` inside with `break` and get rid of the this `func`.

VladoLavor:
Ok but I intentionally didn't split it because I saw no point in it. The example is already very simple and does only basic things. I could remove all those checks which add lines, but I would like to be sure that it actually works. For example, we split Kafka because it was more than twice as long, but I don't think the example is huge and with splitting it we get a lot of duplicated code.

examples/process-manager-plugin/test-process/test-process.go
VladoLavor: > Perhaps just get rid of the TestApp altogether if the goal is to have very simple example, 

That would create such an simple example that it does not work anymore. Examples are intended to work with real data, otherwise we can just write some pseudo code to the readme with the same result, which is not as helpful. Yes we want to have examples as simple as possible but I think this is just an oversimplification.

ondrej-fabry:
It's an example, we don't care about duplicated code there at all. Examples should provide the smallest possible piece of code that does single thing. That way you don't confuse new users when they try to find out how to use some feature. Over 450 lines of code is just too much for examples.

PR #358, opened 2018-11-26
Support non-fatal errors in status check

lukasmacko:


health/probe/plugin_impl_probe.go
```diff
@@ -101,8 +113,8 @@ func (p *Plugin) readinessProbeHandler(formatter *render.Render) http.HandlerFun
 // livenessProbeHandler handles k8s liveness probe.
 func (p *Plugin) livenessProbeHandler(formatter *render.Render) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
-		stat := p.StatusCheck.GetAgentStatus()
-		statJSON, _ := json.Marshal(p.StatusCheck.GetAgentStatus())
+		stat := p.getAgentStatus()
+		statJSON, _ := json.Marshal(stat)
```
VladoLavor: any reason to ignore error here?

health/probe/plugin_impl_probe.go
```diff
@@ -113,3 +125,41 @@ func (p *Plugin) livenessProbeHandler(formatter *render.Render) http.HandlerFunc
...
+				}
+			}
+
+		}
+	}
+
+	return exposedStatus
+}
+
+func isInSlice(haystack []string, needle string) bool {
```
VladoLavor: I would rather use some context-related naming

health/probe/plugin_impl_probe.go
lukasmacko: It was ignored before and I'm not sure whether it's worth to log it or do something about it. Any suggestion?

health/probe/plugin_impl_probe.go
lukasmacko: similar function https://golang.org/pkg/sort/#SearchInts uses single letter names, should I rename it?

PR #361, opened 2018-12-06
BoltDB watcher

ondrej-fabry:
This PR implements Watch for BoltDB plugin.

### DIFF
--- a/db/keyval/bolt/bolt.go
+++ b/db/keyval/bolt/bolt.go
@@ -16,9 +16,11 @@ package bolt
 
 import (
 	"bytes"
-	"errors"
+	"context"
 	"fmt"
 	"os"
+	"strings"
+	"sync"
 
 	"github.com/boltdb/bolt"
 	"github.com/ligato/cn-infra/datasync"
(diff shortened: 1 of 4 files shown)


db/keyval/bolt/bolt.go
```diff
@@ -95,30 +91,69 @@ func (client *Client) GetValue(key string) (data []byte, found bool, revision in
 	return data, found, 0, err
 }
 
+// Put stores given data for the key
+func (c *Client) Put(key string, data []byte, opts ...datasync.PutOption) (err error) {
+	boltLogger.Debugf("Put: %q (len=%d)", key, len(data))
```
VladoLavor: Maybe the boltLogger could be inside Client instead of a global variable.

ondrej-fabry: It's defined as global variable:
- to be able to log even outside of BoltDB client instance 
- to make clear distinction between normal plugin loggers (used for logging plugin related logs) and this low-level client loggers 
These kind of loggers are expected to be explicitly turned on to debug level by environmental variable and not be influenced by `INITIAL_LOGLVL`. Otherwise this would do excessive amounts of unnecessary logging for most of the general use at time when the BoltDB client is not being debugged.

PR #364, opened 2018-12-12
plugin manager enhancements and fixes

VladoLavor:
- [x] set PDeathSignal
- [x] remove unnecessary logs from status parser
- [x] introduce status getter
- [x] do not start watcher on process creation
- [x] set new process status
- [x] add option to redirect logs/errors to custom writer
- [x] choose better naming for package and APIs
- [x] ability to set env parameter


processmanager/process_impl.go
```diff
@@ -12,57 +12,92 @@
...
+		}
+		if p.options.errWriter != nil {
+			errOut, err := cmd.StderrPipe()
+			if err != nil {
+				p.log.Errorf("failed to get stderr pipe: %v", err)
+			}
+			p.watchOutput(p.options.errWriter, errOut)
+		}
+		// detach
+		if p.options.detach {
```
ondrej-fabry: Shouldn't this include else case with setting of `Pdeathsig`? The **detached process** is _opt-in_ feature. To me, this implicates that we want all **non-detached processed** die with their parent process on crash/panic/kill.

processmanager/process_impl.go
```diff
@@ -12,57 +12,92 @@
...
-	if p.options != nil && p.options.detach {
-		attr.Sys = &syscall.SysProcAttr{
-			Setpgid: true,
-			Pgid:    0,
+
+	// if options are set, adjust command attributes, otherwise set last required fields to prepare the command
+	if p.options == nil {
+		cmd.Args = append([]string{p.cmd})
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Pdeathsig: syscall.SIGKILL,
```
ondrej-fabry: It might more convenient to define this signal as global variable, like: `DefaultPDeathSignal`.

VladoLavor: That's the only `Pdeathsig` signal used in PM plugin. I'm not a fan of global variables, what's the advantage here? The signal itself is already a constant.

ondrej-fabry: To allow users to change it to some other type of signal and not be forced to use SIGKILL.

The `Pdeathsig` does not seem significant enough to introduce special option for it, but to allow users to change it, the quickest way is to define global variable for it.

I definitely agree with you that global variables can sometimes be evil, but use of globals is considered bad practice mostly for uses like: storing state or modifying it from many other packages, which can create mutual dependencies and drastically increase complexity.

However there are various other cases where using global variable is adequate. For example, notice the [stdlib](https://golang.org/pkg/) and how some of the packages utilize global variables:

- [net/http](https://golang.org/pkg/net/http/#pkg-variables)
- [flag](https://golang.org/pkg/flag/#pkg-variables)
- [os](https://golang.org/pkg/os/#pkg-variables)

I personally think that using global variables for default values (os) is very clean way to allow users to modify it if needed. Or using globals for default instances (net/http) to allow easier access to instances with sane defaults.

processmanager/process_impl.go
```diff
@@ -12,57 +12,96 @@
...
-		Files: []*os.File{os.Stdin, nil, nil},
-	}
-	// Syscall if process should be detached from parent
-	if p.options != nil && p.options.detach {
-		attr.Sys = &syscall.SysProcAttr{
-			Setpgid: true,
-			Pgid:    0,
+
+	// if options are set, adjust command attributes, otherwise set last required fields to prepare the command
+	if p.options == nil {
```
ondrej-fabry: It seems that these if cases have redundant code. You could introduce variable `defaultOptions`, containing these options from first if case. Then the first case can be removed and default options will be set by the other one.

PR #366, opened 2019-01-04
plugin lookup resolves slice-type fields

VladoLavor:
Plugin lookup does not process slice-type fields (or interfaces where the inner type is slice). Such a field was skipped and all potential plugins inside were ignored, which is causing some cases where the plugin order is still important because the plugin dependency order is not correct.

Example:
VPP plugin defines Publishers with type KeyProtoValWriter, which is an interface representing a list of kvdbsync plugins. Since those plugins were not included, the top level plugin has to define ETCDDataSync (or any other data sync) before the VPP plugin, otherwise, the resolution will fail on unmet dependencies.

This change also causes that the loading order of status check plugin and data sync plugins is reverted (status check defines KeyProtoValWriter, which was skipped before), so the registration was moved to AfterInit().

### DIFF
--- a/agent/README.md
+++ b/agent/README.md
@@ -1,7 +1,7 @@
 # Agent [![GoDoc](https://godoc.org/github.com/ligato/cn-infra/agent?status.svg)](https://godoc.org/github.com/ligato/cn-infra/agent)
 
-The **agent** package provides life-cycle managment agent for plugins.
-It intented tp be used as a base point of your program used in main package.
+The **agent** package provides the life-cycle management agent for plugins.
+It is intended to be used as a base point of an application used in main package.
 
 ```go
 func main() {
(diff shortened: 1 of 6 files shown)


milanlenco:
Don't forget to cherry-pick also to dev branch, otherwise LGTM.

ondrej-fabry:
> Plugin lookup does **not** process slice-type fields
This confused me at first.. but seems like `not` was put there by mistake?

VladoLavor:
> This confused me at first.. but seems like not was put there by mistake?
That was meant as the current state

ondrej-fabry:
Oh right. Perhaps, in the future it'd be better to use *past tense* or include word like *currently* to make it clear that it describes an old behavior.

agent/README.md
```diff
@@ -18,14 +18,15 @@ func main() {
...
+- `QuitOnClose(chan)` sets signal used to quit the running agent when closed
+- `QuitSignals(signals)` sets signals used to quit the running agent (default: SIGINT, SIGTERM)
 - `StartTimeout(dur)/StopTimeout(dur)` sets start/stop timeout (defaults: 15s/5s)
-- add plugins to list of plugins managed by agent using:
-  - `Plugins(...)` adds just single plugins
-  - `AllPlugins(...)` adds plugin along with all of its plugin deps
+
+Add plugins to list of plugins managed by agent using:
+- `Plugins(...)` adds just single plugins
+- `AllPlugins(...)` adds plugin along with all of its plugin deps
```
ondrej-fabry: Since the plugin lookup seems to get more complicated now. It might be good idea to document all the possible situations which are supported by the lookup.

Currently the lookup will search for plugin dependencies in:
- all non-nil fields of a plugin that implement `infra.Plugin` interface
- including fields from all anonymous fields recursively
- including elements of from slice fields

VladoLavor: Agree, and want to prepare the new documentation/guide for the plugin lookup where this will be mentioned it more detail. I plan it as a new PR, so let's omit it here for now if you don't mind.

PR #380, opened 2019-03-12
bolt plugin: add option to filter duplicate notifications

milanlenco:


db/keyval/bolt/plugin.go
```diff
@@ -25,9 +25,10 @@ import (
 
 // Config represents configuration for Bolt plugin.
 type Config struct {
-	DbPath      string        `json:"db-path"`
-	FileMode    os.FileMode   `json:"file-mode"`
-	LockTimeout time.Duration `json:"lock-timeout"`
+	DbPath          string        `json:"db-path"`
+	FileMode        os.FileMode   `json:"file-mode"`
+	LockTimeout     time.Duration `json:"lock-timeout"`
+	FilterDupNotifs bool          `json:"filter-duplicate-notifications"`
```
VladoLavor: Do we need an option to allow duplicated notifications? When it can be useful?

milanlenco: It was meant to allow to avoid calling bytes.Equal for every change in cases when it is known there will be no duplicates or when they are not harmful for user code. But I'm also undecided whether to have this option or not.

PR #393, opened 2019-05-13
supervisor plugin

VladoLavor:
The new bootstrap plugin specifying a config file with a list of process definitions serves as a process manager plugin extension. Process definition is passed to the manager which handles the process lifecycle.

Bootstrap config options:
- define process name and binary file
- define a log file where the output will be written
- start process with arguments
- set a list of other processes automatically terminated if the given instance was stopped


processes/bootstrap/plugin.go
```diff
@@ -0,0 +1,224 @@
...
+		if _, err := instance.StopAndWait(); err != nil {
+			p.Log.Errorf("bootstrap close error: failed to stop process %s: %v", name, err)
+		}
+	}
+
+	return nil
+}
+
+// String is a plugin name
+func (p *Plugin) String() string {
```
ondrej-fabry: No need for this, it's provided by `infra.PluginDeps`.

processes/bootstrap/plugin.go
```diff
@@ -0,0 +1,224 @@
...
+			p.Log.Errorf("unable to start process %s: %v", process.Name, err)
+			continue
+		}
+		processStat := make(chan status.ProcessStatus)
+		pLogger := newPmLogger(process.Name, process.LogFilePath)
+		instance := p.pm.NewProcess(process.Name, process.BinaryPath, pm.Args(process.Args...),
+			pm.Writer(pLogger, pLogger), pm.Notify(processStat), pm.AutoTerminate())
+		if err := instance.Start(); err != nil {
+			p.Log.Errorf("Error starting process %s: %v", process.Name, err)
+			os.Exit(-1)
```
ondrej-fabry: Plugins should not exit process. The `start` function should return error and let caller handle it.

processes/bootstrap/plugin.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+	}
+	delete(p.instances, name)
+	if len(p.instances) == 0 {
+		p.Log.Info("No more processes are running, exiting")
+		os.Exit(0)
+	}
+}
+
+
```
ondrej-fabry: Extra newlines.

processes/bootstrap/plugin.go
```diff
@@ -0,0 +1,224 @@
...
+type Bootstrap interface {
+	// GetProcessNames returns names of all running process instances
+	GetProcessNames() []string
+	// GetProcessByName returns an instance of given process
+	GetProcessByName(name string) pm.ProcessInstance
+}
+
+// Plugin is a bootstrap plugin structure
+type Plugin struct {
+	mx sync.Mutex
```
ondrej-fabry: The mutexes should generally be declared next to variables they protect, in this case just before `instances` variable.

processes/bootstrap/plugin.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// String is a plugin name
+func (p *Plugin) String() string {
+	return p.PluginName.String()
+}
+
+// GetProcessNames returns names of all running process instances
+func (p *Plugin) GetProcessNames() (names []string) {
+	for name := range p.instances {
```
ondrej-fabry: This should be protected by mutex.

processes/bootstrap/plugin.go
```diff
@@ -0,0 +1,224 @@
...
+		names = append(names, name)
+	}
+	return names
+}
+
+// GetProcessByName returns an instance of given process
+func (p *Plugin) GetProcessByName(reqName string) pm.ProcessInstance {
+	for name, instance := range p.instances {
```
ondrej-fabry: This should be protected by mutex.

processes/bootstrap/plugin.go
```diff
@@ -0,0 +1,224 @@
...
+		return errors.Errorf("name %s already exists", process.Name)
+	}
+	if process.BinaryPath == "" {
+		return errors.Errorf("binary %s path not defined", process.Name)
+	}
+	return nil
+}
+
+func (p *Plugin) watch() {
+	for name, instance := range p.instances {
```
ondrej-fabry: This should be protected by mutex.

processes/bootstrap/plugin.go
VladoLavor: Good point

processes/bootstrap/plugin.go
```diff
@@ -0,0 +1,224 @@
...
+
+// Deps are bootstrap dependencies (standard infra plugins + process manager)
+type Deps struct {
+	pm pm.ProcessManager
+	infra.PluginDeps
+}
+
+// Init starts processes and their watchers defined by the configuration file
+func (p *Plugin) Init() error {
+	p.instances = make(map[string]pm.ProcessInstance)
```
ondrej-fabry: This should be protected by mutex.

processes/bootstrap/plugin.go
```diff
@@ -0,0 +1,224 @@
...
+				p.Log.Errorf("Attempt to stop %s failed: %v", toStop, err)
+			}
+		} else {
+			p.Log.Info("Process %s is no longer running", toStop)
+		}
+	}
+	delete(p.instances, name)
+	if len(p.instances) == 0 {
+		p.Log.Info("No more processes are running, exiting")
+		os.Exit(0)
```
ondrej-fabry: Plugins should not exit process.

processes/bootstrap/pmlogger.go
```diff
@@ -0,0 +1,79 @@
...
+	pmLog.log.SetLevel(logging.DebugLevel)
+	return pmLog
+}
+
+// Write to logfile and print given message
+func (l pmLogger) Write(p []byte) (n int, err error) {
+	mx.Lock()
+	defer mx.Unlock()
+
+	file, err := os.OpenFile(l.logfilePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
```
ondrej-fabry: We definitely do not want to open and close file for each write. That would be really expensive performance-wise.

ondrej-fabry: Files should be opened ONCE before program is started and closed after it exists. Also, the log should not be written to the file directly, instead the file should be wrapper with `bufio.NewWriter` to handle buffering and not execute writes for every byte.

processes/bootstrap/pmlogger.go
```diff
@@ -0,0 +1,79 @@
...
+	"io"
+	"log"
+	"os"
+	"sync"
+
+	"github.com/ligato/cn-infra/logging"
+	"github.com/ligato/cn-infra/logging/logrus"
+)
+
+const defaultLogfilePath = "/var/log/vpp-agent-init.log"
```
ondrej-fabry: Not a good default log name. There should be only default log path and name should be derived from process name.

processes/bootstrap/pmlogger.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return 0, errors.Errorf("error opening file: %v", err)
+	}
+	defer func() {
+		if err := file.Close(); err != nil {
+			l.log.Errorf("error closing file %s", l.logfilePath)
+		}
+	}()
+
+	writer := io.Writer(file)
```
ondrej-fabry: This does not really do anything, since the [`io.Writer`](https://golang.org/pkg/io/#Writer) is just interface.

processes/bootstrap/pmlogger.go
```diff
@@ -0,0 +1,79 @@
...
+	log.SetOutput(writer)
```
ondrej-fabry: This would not be very efficient, since [`log.SetOutput`](https://golang.org/pkg/log/#SetOutput) changes output destination for the standard logger. You can just create the Logger instance with `log.NewLogger` for each process to prefix their output.

processes/bootstrap/bootstrap.conf
```diff
@@ -0,0 +1,10 @@
+processes:
+  - name: "vpp"
+  log-file-path: "/tmp/bootstrap.log"
```
ondrej-fabry: This is not valid YAML. Try using some online validator, it throws: `Error: You cannot define a mapping item when in a sequence`.

The lines 3-6 and 8-10 should be indented together with `name`.

examples/process-manager-plugin/bootstrap/main.go
```diff
@@ -0,0 +1,106 @@
...
+				Name:           "p3",
+				LogFilePath:    "example.log",
+				BinaryPath:     "../test-process/test-process",
+				TriggerStopFor: []string{"p2"},
+			},
+		},
+	}
+
+	// start plugin
+	bsp := bs.NewPlugin(bs.UseConf(*conf))
```
ondrej-fabry: You could declare `conf` as value (not pointer) and avoid changing to value here.

examples/process-manager-plugin/bootstrap/main.go
```diff
@@ -0,0 +1,106 @@
...
+
+	go func() {
```
ondrej-fabry: You could actually call `a.Start()` together with `defer a.Stop()` here and avoid using goroutine and sleep.

processes/bootstrap/config.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// Path to the process binary file
+	BinaryPath string `json:"binary-path"`
+
+	// A list of arguments the process will be started with
+	Args []string `json:"args"`
+
+	// This parameter allows to define another processes, which will be stopped
+	// when the current instance is terminated (processes are defined by name).
+	TriggerStopFor []string `json:"trigger-stop-for"`
```
ondrej-fabry: The `TriggerStopFor` has very narrow use-case and doesn't allow a lot of flexibility for users. It also doesn't fully mirror the functionality provided by `supervisord_kill.py` in vpp-agent (I assume that's why `TriggerStopFor` was introduced here), because this stops other specific process defined for this plugin instead of exiting it. And by adding any other process to the list of processes would break the container-exit functionality (if it wasn't also _chained_ the same way).

To provide more generic behaviour that could be completely flexible and extensible by users I would suggest to remove `TriggerStopFor` field from `Process` and add `Hooks` field of type `[]Hook` to `Config` which could provide arbitrary hooks for this plugin. These hooks could listen to some defined events and execute some external command (script) which would receive attributes of event as environmental variables. This would provide very similar functionality to what supervisord currently provides via [`[eventlistener:x]`](http://supervisord.org/configuration.html#eventlistener-x-section-values) and could be used in any way user wants.

milanlenco:
> I propose to call this plugin **`Launcher`** or **`Supervisor`**, since it launches processes and supervise them.

Personally I like `supervisor` more... (it will be clear what this is an alternative for)

> And for the directory that now contains this plugin and processmanager plugin and I think we should stick with something shorted, and preferably also something that would avoid stutter for processmanager plugin, for example:
> 
> * `proc` (`proc/processmanager`, `proc/launcher`)
> * `exec` (`exec/processmanager`, `exec/launcher`)

With `proc` maybe we could shorten to `proc/manager`? But actually I also like `exec` more and then leaving processmanager as is (?)

processes/bootstrap/bootstrap.conf
VladoLavor: IDE aligned it when copied

processes/bootstrap/config.go
VladoLavor: > adding any other process to the list of processes would break the container-exit functionality
I see your point, we can just introduce some specific keyword here (or adjust the field) so not all processes need to be listed here.

>  These hooks could listen to some defined events and execute some external command (script) which would receive attributes of an event as environmental variables
I thought we want to get rid of scripts, that's why I found this functionality unnecessary here.

VladoLavor:
Let's call it `supervisor` and rename the folder to `exec`.

exec/supervisor/plugin.go
```diff
@@ -0,0 +1,211 @@
...
+func (p *Plugin) Init() error {
+	p.instances = make(map[string]pm.ProcessInstance)
+
+	if err := p.getConfig(); err != nil {
+		return err
+	}
+
+	// run watchers even if there are failed processes. They will be handled
+	// as terminated and exit immediately
+	err := p.start(p.config.Processes)
```
ondrej-fabry: We don't want to start watch if there is an error. Check error before calling watch.

exec/supervisor/plugin.go
```diff
@@ -0,0 +1,211 @@
...
+			continue
+		}
+		processStat := make(chan status.ProcessStatus)
+		pLogger, err := pmLogManager.newPmLogger(process.Name, process.LogFilePath)
+		if err != nil {
+			p.Log.Errorf("error preparing process logger %s: %v", process.Name, err)
+			inError = append(inError, process.Name)
+			continue
+		}
+		instance := p.pm.NewProcess(process.Name, process.BinaryPath, pm.Args(process.Args...),
```
ondrej-fabry: It'll be more readable to put each poption on separate line.

exec/supervisor/supervisor.conf
```diff
@@ -0,0 +1,10 @@
+processes:
```
ondrej-fabry: Shouldn't all these lines be commented out? AFAIR, these represent some default config value, no?

exec/supervisor/plugin.go
VladoLavor: The watch is not started (I should rewrite the comment), the process is handled as terminated. Another option is to stop processes which were previously started before (it there were any) and return the error.

exec/supervisor/supervisor.conf
VladoLavor: Can be, since there is hard to define a default config - everyone has to put its own values here.

exec/supervisor/config.go
```diff
@@ -0,0 +1,80 @@
...
+// Process is a single process instance started by the supervisor
+type Process struct {
+	// Unique name of the process
+	Name string `json:"name"`
+
+	// File where the log output will be written
+	LogFilePath string `json:"log-file-path"`
+
+	// Path to the process binary file
+	BinaryPath string `json:"binary-path"`
```
ondrej-fabry: This should rather be named `Executable`, since the actual executed command does not necessarily have to be "binary", it could be simple bash script - text file.

processes/bootstrap/config.go
ondrej-fabry: We wanted to get rid of python dependency. Simple bash script is harmless.

exec/supervisor/config.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// A list of arguments the process will be started with
+	Args []string `json:"args"`
+
+	// Marks process as required. A required process triggers termination
+	// or all other processes known to supervisor
+	Required bool `json:"required"`
+
+	// Hooks allow to run custom commands/events at process termination
+	Hooks []Hook `json:"hooks"`
```
ondrej-fabry: In [the previous review comment](https://github.com/ligato/cn-infra/pull/393/#discussion_r283325304), I actually proposed to add `Hooks` to the `Config` type, instead of `Process`. 

This allows to have single global hook that receives events for all programs. The event info can be passed to the hook command using env vars:

- `SUPERVISOR_EVENT_TYPE` - represents type of hook event; for now only one type: `process-state` ([more](http://supervisord.org/events.html#event-types) could be added later)
- `SUPERVISOR_PROCESS_NAME` - value of `Process.Name`
- `SUPERVISOR_PROCESS_STATE` - value of process status coming from notification channel

These are enough for handling process state changes for any process. We can add more later when needed.

exec/supervisor/plugin.go
```diff
@@ -0,0 +1,222 @@
...
+	var process Process
+	for _, processCfg := range p.config.Processes {
+		if name == processCfg.Name {
+			process = processCfg
+			break
+		}
+	}
+
+	// execute hooks
+	for _, hook := range process.Hooks {
```
ondrej-fabry: We don't want to execute hooks only when terminating program. The hooks should be executed mainly in `watchProcess` for now to handle any change of process state.

exec/supervisor/plugin.go
```diff
@@ -0,0 +1,222 @@
...
+		if hook.Command != "" {
+			if _, err := exec.Command(hook.Command, hook.CmdArgs...).Output(); err != nil {
```
ondrej-fabry: Don't forget about stderr.

exec/supervisor/pmlogger.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/ligato/cn-infra/logging"
+	"github.com/ligato/cn-infra/logging/logrus"
+	"github.com/pkg/errors"
+)
+
+const (
+	defaultLogfilePath = "/var/log/"
+
+	logfileSuffix = "-file"
```
ondrej-fabry: Why not just `.log` as commonly used for log files.

examples/process-manager-plugin/supervisor/main.go
```diff
@@ -0,0 +1,110 @@
...
+		panic(err)
+	}
+	defer func() {
+		if err := a.Stop(); err != nil {
+			panic(err)
+		}
+	}()
+
+	// give the agent time to start
+	time.Sleep(5 * time.Second)
```
ondrej-fabry: This should not be needed. The `agent.Start()` will return only after all plugins are successfully initialized.

exec/supervisor/config.go
```diff
@@ -0,0 +1,80 @@
...
+import "github.com/pkg/errors"
+
+// Config represents configuration file for the supervisor procedure
+type Config struct {
+	// The configuration file consists from a list of processes
+	Processes []Process `json:"processes"`
+}
+
+// Process is a single process instance started by the supervisor
+type Process struct {
```
ondrej-fabry: I think this should rather be named `Program` to avoid disambiguation with `processmanager.Process`.

exec/supervisor/plugin.go
```diff
@@ -0,0 +1,222 @@
...
+	if len(inError) != 0 {
+		return errors.Errorf("following processes end up with error: %v", inError)
+	}
+	return nil
+}
+
+func (p *Plugin) validate(process *Process) error {
+	if process.Name == "" {
+		// so it will not be logged as empty space
+		process.Name = unnamedProcess
```
ondrej-fabry: Why not rather derive the name from `BinaryPath` field? It would allow multiple programs without name, which the current implementation doesn't support and will fail.

exec/supervisor/plugin.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+}
+
+func (p *Plugin) handleTerminated(name string, instance pm.ProcessInstance) {
+	p.mx.Lock()
+	defer p.mx.Unlock()
+
+	// find original config
+	var process Process
+	for _, processCfg := range p.config.Processes {
```
ondrej-fabry: Generally, when naming variables you should use shorter names in smaller scopes and longer verbose names in larger scopes (loops -> funcs -> packages). Referring to variables `process` and `processCfg` here.

exec/supervisor/plugin.go
```diff
@@ -0,0 +1,222 @@
...
+	for name, instance := range p.instances {
+		if name == reqName {
+			return instance
+		}
+	}
+	return nil
+}
+
+func (p *Plugin) start(processes []Process) error {
+	pmLogManager := newPmLoggerManager()
```
ondrej-fabry: The entire logging for programs is totally over-engineered. Adding to my [previous review comment](https://github.com/ligato/cn-infra/pull/393/#discussion_r283307806):

https://github.com/ligato/cn-infra/blob/8933a15080a07ab92ae50bd8ccab73da5455f97f/exec/supervisor/pmlogger.go#L57

There is no need for logs to go through logrus, the [`log.Logger`](https://golang.org/pkg/log/#Logger) from stdlib is enough for our use-case, even that seems unnecessary when logging to actual file. 

https://github.com/ligato/cn-infra/blob/8933a15080a07ab92ae50bd8ccab73da5455f97f/exec/supervisor/pmlogger.go#L110

This variable is useless. Just use the log from Plugin.

https://github.com/ligato/cn-infra/blob/8933a15080a07ab92ae50bd8ccab73da5455f97f/exec/supervisor/pmlogger.go#L38

There is no need for channels. It's okay to open file more than once when using it in multiple programs.

https://github.com/ligato/cn-infra/blob/8933a15080a07ab92ae50bd8ccab73da5455f97f/exec/supervisor/pmlogger.go#L42

This is not even called and even if it was, it might panic because it's not closed by [the producer](https://github.com/ligato/cn-infra/blob/8933a15080a07ab92ae50bd8ccab73da5455f97f/exec/supervisor/pmlogger.go#L124).

---

And even with all of this code for logging it does not support the simple use-case which we currently have in vpp-agent, which is: if log file is not defined, use stdout/stderr or program that runs supervisor. We want to keep this behaviour and we actually want to ensure that the output printed by child processes (agent in our case) is printed with color. This should return `true` in the agent process when using supervisor.

VladoLavor:
rewritten, will be introduced in another PR

PR #396, opened 2019-05-30
supervisor terminates programs before closing

VladoLavor:

### DIFF
--- a/exec/supervisor/plugin.go
+++ b/exec/supervisor/plugin.go
@@ -42,8 +42,9 @@ type Plugin struct {
 	mx       sync.Mutex
 	programs map[string]*processWithStateChan
 
-	// common channel where all executed programs send their process state
-	hookChan chan *processEvent
+	// hook channel where all executed programs send their process state
+	hookEventChan chan *processEvent
+	hookDoneChan  chan struct{}
 
 	// supervisor configuration
 	config *Config
@@ -88,7 +89,8 @@ func (p *Plugin) Init() error {
 		return errors.Errorf("supervisor config file not defined or does not contain any program")
 	}
 	p.programs = make(map[string]*processWithStateChan)
-	p.hookChan = make(chan *processEvent)
+	p.hookEventChan = make(chan *processEvent)
+	p.hookDoneChan = make(chan struct{})
 
 	go p.watchEvents()
 	// start programs in another go routine (do not block init since it may take a while)
@@ -114,7 +116,10 @@ func (p *Plugin) Close() error {
 
 	// close hook watcher when all programs are terminated
 	p.wg.Wait()
-	close(p.hookChan)
+	close(p.hookEventChan)
+
+	// wait for hook watcher to finish
+	<-p.hookDoneChan
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


exec/supervisor/plugin.go
```diff
@@ -179,24 +194,33 @@ func (p *Plugin) execute(program *Program) error {
...
+		doneChan:  doneChan,
 		svLogger:  svLogger,
 	}
 
 	return nil
 }
 
-func (p *Plugin) watch(stateChan chan status.ProcessStatus, name string) {
+func (p *Plugin) watch(stateChan chan status.ProcessStatus, doneChan chan struct{}, name string) {
+	p.wg.Add(1)
```
ondrej-fabry: I've mentioned this few times before, but to be more specific I will quote important part of documentation for [`WaitGroup.Add`](https://godoc.org/sync#WaitGroup.Add): 

> Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for...

exec/supervisor/plugin.go
```diff
@@ -97,11 +100,22 @@ func (p *Plugin) Init() error {
...
 		if err := program.svLogger.Close(); err != nil {
 			p.Log.Errorf("failed to close logger: %v", err)
 		}
+		// terminate watcher for given supervisor process
+		close(program.doneChan)
 	}
+
+	// close hook watcher when all programs are terminated
+	p.wg.Wait()
 	close(p.hookChan)
```
ondrej-fabry: This is improper handling of channel as I've explained in other PR: https://github.com/ligato/vpp-agent/pull/1318#discussion_r288880877

The `Close` method is called from different goroutine than the goroutine that handles `watch` method and sends events to the `hookChan`.

VladoLavor: Indeed, that's the reason for the wait group. There is a `watch` goroutine for every process started and the hook channel is shared between them. The wait group ensures that no more watchers are running, and then it's safe to close the channel even from a different goroutine.

exec/supervisor/plugin.go
VladoLavor: Agree, but the reason for this rule is to prevent calling `Wait()` before the first `Add()` which may happen when the goroutine execution is delayed.

exec/supervisor/plugin.go
ondrej-fabry: But it doesn't really make much sense why you want to close the `hookChan` when immediately after closing it you return from `Close`.. which means that it will not wait for `watchEvents` to finish anyway so why bother?

exec/supervisor/plugin.go
ondrej-fabry: That is just wrong. You should realize the flow of the plugin. You have chain of goroutines starting from `Init` method and continuing `startPrograms` and then this `watch` where you can never ensure the the time of execution. The `Close` could be called immediately after `Init` when the `p.programs` is still empty, which can easily happen, because of so many unsynchronized goroutines, which would actually make the execution skip to the `Wait` call and immediately exiting, possibly creating a lot of race conditions here. 

This is simply bad practice that breeds unreliable code.

Issue #399, opened 2019-07-24
Support duration format in config files


ondrej-fabry:
Make it possible to define duration as `mydur: 5s` instead of `mydur: 5000000000` for fields with type `time.Duration`.

ondrej-fabry:
Fixed

PR #408, opened 2019-09-30
feat: Support string duration format in config files

rewenset:
Issue #399 

This PR allows setting duration fields, which have `time.Duration` type, in config files not only in nanoseconds:
```yaml
reply-timeout: 5000000000
```
but as a string too
```yaml
reply-timeout: 5s
```


rewenset:
So if I change for example in ETCD `config.go` this field

https://github.com/ligato/cn-infra/blob/9bb0e395ce981f921420130be6d13a1582ed99f1/db/keyval/etcd/config.go#L34

to be not `time.Duration`, but `config.Duration` (which is newly created type introduced in this PR)
then I'll need to find all places this field is used and add `.Duration`, e.g.:
```diff
        dialTimeout := defaultDialTimeout
-       if yc.DialTimeout != 0 {
-               dialTimeout = yc.DialTimeout
+       if yc.DialTimeout.Duration != 0 {
+               dialTimeout = yc.DialTimeout.Duration
        }
```

For now I didn't find better solution.

@ondrej-fabry please advise

config/parser.go
```diff
@@ -33,8 +36,37 @@ func ParseConfigFromYamlFile(path string, cfg interface{}) error {
...
+	err := yaml.Unmarshal(b, &data)
+	if err != nil {
+		return err
+	}
 
-	err = yaml.Unmarshal(b, cfg)
+	dc := &mapstructure.DecoderConfig{
+		DecodeHook: func(in, out reflect.Type, data interface{}) (interface{}, error) {
+			// Only intended to help with cases when string must be set to `time.Duration`
+			if in.Kind() != reflect.String || out.Name() != "Duration" {
```
ondrej-fabry: Checking by type name is not very certain. The [`reflect.Type`](https://golang.org/pkg/reflect/#Type) values are comparable:

```go
reflect.TypeOf(out) == reflect.TypeOf(time.Duration(0))
```
> [Play](https://play.golang.org/p/pI1btkA6uEO)

PR #411, opened 2019-10-23
feat: support for loggerV2

VladoLavor:
Support for GRPC loggingv2, allowing to set logger verbosity. 

### DIFF
--- a/logging/log_api.go
+++ b/logging/log_api.go
@@ -48,6 +48,12 @@ func Warn(args ...interface{}) { DefaultLogger.Warn(args...) }
 // Warnf is for logging with default logger.
 func Warnf(format string, args ...interface{}) { DefaultLogger.Warnf(format, args...) }
 
+// Warning is for logging with default logger.
+func Warning(args ...interface{}) { DefaultLogger.Warning(args...) }
+
+// Warningf is for logging with default logger.
+func Warningf(format string, args ...interface{}) { DefaultLogger.Warningf(format, args...) }
+
 // Error is for logging with default logger.
 func Error(args ...interface{}) { DefaultLogger.Error(args...) }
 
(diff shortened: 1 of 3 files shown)


logging/log_api.go
```diff
@@ -42,11 +42,11 @@ func Info(args ...interface{}) { DefaultLogger.Info(args...) }
 // Infof is for logging with default logger.
 func Infof(format string, args ...interface{}) { DefaultLogger.Infof(format, args...) }
 
-// Warn is for logging with default logger.
```
ondrej-fabry: Could this exported function keep the short version for warning level? Or is there some reasoning behind it?

logging/logrus/entry.go
```diff
@@ -150,10 +151,10 @@ func (entry *Entry) Printf(format string, args ...interface{}) {
 }
 
 // Warnf logs a message at level Warn on the standard logger.
+//
+// Deprecated: use Warningf
 func (entry *Entry) Warnf(format string, args ...interface{}) {
-	if entry.lgEntry.Logger.Level >= lg.WarnLevel {
-		entry.lgEntry.WithFields(entry.fields).Warnf(format, args...)
-	}
+	entry.Warningf(format, args...)
```
ondrej-fabry: Calling `Warningf` from `Warn` will include additional level in the stack trace and report wrong line for `loc` tag.

logging/log_api.go
```diff
@@ -147,8 +157,8 @@ func (level LogLevel) String() string {
 		return "fatal"
 	case ErrorLevel:
 		return "error"
-	case WarnLevel:
-		return "warn"
+	case WarningLevel:
```
ondrej-fabry: To avoid breaking changes and be more consistent with [logrus levels](https://godoc.org/github.com/sirupsen/logrus#Level) I would suggest to not change this.

PR #413, opened 2019-10-23
feat: enable grpc plugin logging

VladoLavor:
* Enable GRPC logging
* Additional GRPC transport messages can be enabled with `extended-logging` set to `true` in the config file (these are disabled by default since there is plenty of them)

### DIFF
--- a/examples/grpc-plugin/grpc-server/main.go
+++ b/examples/grpc-plugin/grpc-server/main.go
@@ -25,9 +25,6 @@ const PluginName = "myPlugin"
 func main() {
 	grpcPlug := grpc.NewPlugin(
 		grpc.UseHTTP(&rest.DefaultPlugin),
-		grpc.UseConf(grpc.Config{
-			Endpoint: "localhost:9111",
-		}),
 		grpc.UseAuth(&grpc.Authenticator{
 			Username: "testuser",
 			Password: "testpwd",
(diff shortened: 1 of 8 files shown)


examples/grpc-plugin/grpc-server/main.go
```diff
@@ -25,9 +25,6 @@ const PluginName = "myPlugin"
 func main() {
 	grpcPlug := grpc.NewPlugin(
 		grpc.UseHTTP(&rest.DefaultPlugin),
-		grpc.UseConf(grpc.Config{
-			Endpoint: "localhost:9111",
```
ondrej-fabry: The `grpc.UseConf` was used as an example here. Perhaps you could enable extended logging here instead of via config file?

VladoLavor: That is also an option. I followed the [readme ](https://github.com/ligato/cn-infra/blob/master/examples/grpc-plugin/README.md) which gives an option to use the config file, however, it is ignored when used together with `UseConf`.

ondrej-fabry: Yes, that is to provide a way to programmatically override config files.

ondrej-fabry:
Fix travis failure and I'll merge

examples/grpc-plugin/grpc-server/main.go
VladoLavor: It would be a good idea to detect situations the config file was set programmatically and also from outside and inform about it in the log. One may not realize it and then it looks like some weird bug.

PR #424, opened 2020-03-27
feat: Rate limiting

ondrej-fabry:


rpc/grpc/plugin_impl_grpc.go
```diff
@@ -86,31 +89,48 @@ func (p *Plugin) Init() (err error) {
...
 
-		// get server options from config
-		opts := p.Config.getGrpcOptions()
+		var unaryChain []grpc.UnaryServerInterceptor
+		var streamChain []grpc.StreamServerInterceptor
 
-		// add custom server options
-		opts = append(opts, p.serverOpts...)
+		// Rate limiting middleware
+		if p.limiter == nil {
```
rastislavszabo: Does this mean that the limiter will be enabled by default, with rate and burst size hardcoded in `defaultRateLimiter()`? I'm not sure this is a good idea, people unaware of the limiter may get confused. I would rather have it disabled by default, and configurable (enabled/disabled, rate & burst size) via config file.

ondrej-fabry: I guess you are right, will fix.

rpc/rest/auth.go
```diff
@@ -20,7 +20,7 @@ import (
 	"strings"
 )
 
-func auth(h http.Handler, auth BasicHTTPAuthenticator) http.HandlerFunc {
+func authMiddleware(h http.Handler, auth BasicHTTPAuthenticator) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		user, pass, _ := r.BasicAuth()
 		if !auth.Authenticate(user, pass) {
```
rastislavszabo: can we have a separate rate limiter for authentication - limiting failed authentication attempts, ideally per user (IP address?)

ondrej-fabry: Yeah, I think we can. I assume both should be configurable separately? 

*Per user* could work for rate limiting API access of logged in users and *per IP* address could work for rate limiting failed logins. What do you think? @rastislavszabo

rastislavszabo: >  I assume both should be configurable separately?

Yeah, auth rate limiting usually will be much more strict

> Per user could work for rate limiting API access of logged in users and per IP address could work for rate limiting failed logins. 

great!

PR #433, opened 2021-02-17
refactor: Drop usage of tlsutils from coreos/etcd

rewenset:
This PR replaces "github.com/coreos/etcd/pkg/tlsutil" with standard library functions.

Also, I executed `go mod tidy` and it removed few lines from go.sum. This is pretty confusing for me... I can revert that if anything.
And the last thing, I've refactored it but not tested. What could possibly go wrong? :)

### DIFF
--- a/db/keyval/etcd/config.go
+++ b/db/keyval/etcd/config.go
@@ -17,12 +17,13 @@ package etcd
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


ondrej-fabry:
> 
> Also, I executed `go mod tidy` and it removed few lines from go.sum. This is pretty confusing for me... I can revert that if anything.
> And the last thing, I've refactored it but not tested. What could possibly go wrong? :)

Running `go mod tidy` sometimes removes unnecessary entries from `go.sum`, for example versions that are no longer used. Read more details here: https://golang.org/ref/mod#go-mod-tidy

ondrej-fabry (approved):

